PR #15, opened 2020-02-10
Implemented ScoreVerifier

artem-smotrakov:
Here is a list of updates:

- Added `AbstractVerifier` and `ScoreVerifier` classes
- Added `ScoreVerifierTest`
- Renamed `RatingVerifier.failedVectors()` to `runImpl()`
- `runImpl()` returns a list instead of a set

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
@@ -1,0 +1,61 @@
+package com.sap.sgs.phosphor.fosstars.model.qa;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public abstract class AbstractVerifier {
+
+  /**
+   * A list of test vectors.
+   */
+  final List<TestVector> vectors;
+
+  /**
+   * A logger.
+   */
+  private final Logger logger = LogManager.getLogger(getClass());
+
+  /**
+   * Initialize a verifier.
+   *
+   * @param vectors A list of test vectors.
+   */
+  AbstractVerifier(List<TestVector> vectors) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Check the rating against the test vectors, and throws a {@link VerificationFailedException} if
+   * at least one test vector failed.
+   *
+   * @throws VerificationFailedException If at least one test vector failed.
+   */
+  public final void run() throws VerificationFailedException {
+    List<FailedTestVector> failedVectors = runImpl();
+    for (FailedTestVector vector : failedVectors) {
```
Sachpat: How about printing the logs in specific verifiers by basically extracting this to a method and then teaking the LOG as a parameter from the specific verifier class?

src/main/java/com/sap/sgs/phosphor/fosstars/model/tuning/CMAESWeightsOptimization.java
```diff
@@ -293,7 +300,7 @@ public double value(double[] parameters) {
     }
 
     /**
-     * @return A number of weights which may be adjusted.
+     * Returns a number of weights which may be adjusted.
```
Sachpat: Returns *the number of weights* maybe? as we are specific about weights here?  :)

artem-smotrakov: I agree.

src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
artem-smotrakov: Good point.

artem-smotrakov:
Thanks for the review @Sachpat . I have addressed the comments.

PR #18, opened 2020-02-12
Implemented a verification procedure for ProjectActivityScore

artem-smotrakov:
Main updates:

- Updated `Rating.allFeatures()` to return a set of features.
- Updated and added several methods to use collections of features.
- Added `ProjectActivityScore.Verification` class.
- Added a `ProjectActivityScoreTestVectors` notebook which defines the test vectors.
- Added `ProjectActivityScoreVerification` test which runs the verification.
- Fixed checkstyle warnings in the modified files.


src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/AbstractRating.java
```diff
@@ -102,19 +113,18 @@ public boolean equals(Object o) {
       return false;
     }
     AbstractRating that = (AbstractRating) o;
-    return Objects.equals(score, that.score) &&
-        version == that.version;
+    return Objects.equals(score, that.score) && version == that.version;
```
Sachpat: Just a question, why don't we use Objects.equals(version, that.version) here?

artem-smotrakov: `version` is a enum, the simplest way to compare them is `==`.

artem-smotrakov:

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/AbstractRating.java
Sachpat: Aah missed that it's an enum :D thanks for the clarification!

PR #21, opened 2020-02-13
Added test vectors for feature-based scores

artem-smotrakov:
This update introduces test vectors for feature-based scores which are used in the open-source security rating:

- `VulnerabilityLifetimeScore`
- `UnpatchedVulnerabilitiesScore`
- `ProjectSecurityTestingScore`
- `ProjectSecurityAwarenessScore`
- `CommunityCommitmentScore`
- `ProjectPopularityScore`

Test vectors for the `ProjectActivityScore` have been added earlier in #18 

`commons.py` contains common functions and classes which are used in the notebooks.


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/CommunityCommitmentScore.java
```diff
@@ -61,4 +69,43 @@ public ScoreValue calculate(Value... values) {
...
+    }
+
+    /**
+     * Creates an instance of {@link Verification} for a specified score. The method loads test
+     * vectors from a default resource.
+     *
+     * @param score The score to be verified.
+     * @return An instance of {@link CommunityCommitmentScore}.
+     */
+    static Verification createFor(CommunityCommitmentScore score) throws IOException {
```
Sachpat: Hmm, just a question. If this method is needed by classes outside this class, shouldn't this be public static?

artem-smotrakov: It's currently used only by classes in the same package. That's why it's package-private. I would like to restrict access as much as I can.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/UnpatchedVulnerabilitiesScoreVerification.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.sap.sgs.phosphor.fosstars.model.qa.VerificationFailedException;
+import java.io.IOException;
+import org.junit.Test;
+
+public class UnpatchedVulnerabilitiesScoreVerification {
+
+  @Test
+  public void verify() throws IOException, VerificationFailedException {
+    UnpatchedVulnerabilitiesScore.Verification.createFor(
```
Sachpat: As discussed, I think we can remove the parameter for createFor() ?

artem-smotrakov: I forgot to mention that the `UnpatchedVulnerabilitiesScore` is not a singleton so that there may be multiple instances of this class. Right now, we use the `Verification` class to verify only `OssScores.UNPATCHED_VULNERABILITIES` but in the future we may want to verify other instances. That's why we may need this parameter in the `createFor()` method.

Sachpat: Ok, makes sense then .

artem-smotrakov:

Issue #26, opened 2020-02-14
Test vectors for score-based scores


artem-smotrakov:
Currently it's possible to add test vectors only for ratings. #14 introduces test vectors for feature-based scores. However, we also have scores which consume other sub-scores. For example, see `OssSecurityScore` which is based on `WeightedCompositeScore`. Such scores needs to be covered by test vectors as well.

PR #27, opened 2020-02-17
Test vectors for score-based scores

artem-smotrakov:
This is a part of #26. Here is a list of main updates:

- Added a features and a value which represent a double.
- All scores become features which hold a double value.
- Updated the `ScoreValue` class to implement the `Value` interface.
- Moved the `ScoreValue` and `RatingValue` classes to the `com.sap.sgs.phosphor.fosstars.model.value` package.
- Updated the `WeightedCompositeScore` class to use pre-calculated scores.
- Fixed `equals()` and `hashCode()` methods.
- Added tests.


src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/DoubleFeature.java
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  @Override
+  public DoubleValue value(Double n) {
+    return new DoubleValue(this, n);
+  }
+
+  @Override
+  public Value<Double> parse(String string) {
+    return value(Double.parseDouble(string));
```
Sachpat: Are you expecting NPEs or NumberFormat Exceptions or you would want to check the string better? )

artem-smotrakov: I would not like to catch this runtime exceptions. I would prefer to let `Double.parseDouble()` throw them according to its javadoc.

Sachpat: No, I mean, do you want to check for the format of the String before atleast for a null? because NPEs doesn't look good in the logs.

artem-smotrakov: I understood what you mean. The `Double.parseDouble()` method expectes a null and throws an NPE if a string is null:

https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#parseDouble(java.lang.String)

I can add an `Objects.requireNotNull(string, ...)` which also throws an NPE if an object is null. The only thing we get here is a custom exception message. If you think it's worth it, I can add such a check. I don't think we should check the format of the string - the `parseDouble()` method is supposed to do that. We can only check for an empty string. If you think it's worth it, I can add such a check. There is no problem.

Sachpat: How about IllegalArgumentException() with the message? :) If you think it's not needed, please go ahead, no worries at all )

src/test/java/com/sap/sgs/phosphor/fosstars/model/feature/DoubleFeatureTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    assertNotEquals(one.hashCode(), two.hashCode());
+    assertEquals(one.hashCode(), anotherOne.hashCode());
+  }
+
+  @Test
+  public void value() {
+    Value<Double> value = new DoubleFeature("test").value(10.2);
+    assertNotNull(value);
+    assertEquals(10.2, value.get(), 0.001);
```
Sachpat: Just to know better, why do we have delta here? and is the delta always positive? :)

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/DoubleFeature.java
artem-smotrakov: I think that a NullPointerException should be thrown if a string is null. IllegalArgumentException can be thrown if a string is empty. I've added the checks.

Sachpat: Ok, thanks for adding it )

PR #28, opened 2020-02-17
Implemented a verification procedure for OssSecurityScore

### LINKED ISSUE #26: Test vectors for score-based scores

artem-smotrakov:

- Added `OssSecurityScore.Verification` class
- Added `OssSecurityScoreTestVectors` notebook which defines the test vectors
- Added `OssSecurityScoreVerification` test which runs the verification
- Updated `loadTestVectorsFromCsvResource()` to expect a set of scores
- Updated `AbstractScore.value()` to return a ScoreValue
- `WeightedCompositeScore` now expects an unknown value for a score
- Updated README and javadoc

This closes #26


src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerification.java
```diff
@@ -108,7 +108,7 @@
 
         String label = record.get("label");
         Label expectedLabel = null;
-        if (!"None".equals(label) && !label.isEmpty() && labelParser != null) {
+        if (!"None".equals(label) && !label.isEmpty()) {
```
Sachpat: Will it be a good idea to extract this "None" value to a global constant and use it everywhere?  Just to counter the fact that spelling differences, case sensitivity etc?

artem-smotrakov: The constant is currently used only in the `AbstractVerification` class. If we start using this constant anywhere else, we'll definitely extract it to a final static field.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScore.java
```diff
@@ -159,20 +160,22 @@ public int hashCode() {
...
+    }
+
+    // if the set of values contains a value for the specified score, then return it
+    if (value instanceof ScoreValue) {
+      return (ScoreValue) value;
     }
 
-    // otherwise, try to calculate a value for the score
-    return score.calculate(values);
+    throw new IllegalArgumentException(String.format(
```
Sachpat: How about adding @throws in the javadoc for this exception?

artem-smotrakov: Yeah, why not.

src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerification.java
Sachpat: Makes sense, how about changing it to equalsIgnoreCase ?

artem-smotrakov: Good point.

PR #33, opened 2020-02-20
Fixed checkstyle issues

artem-smotrakov:
Here is the main updates:

- Fixed checkstyle warnings in in both `src/main/java` and `src/test/java`.
- Updated the checkstyle configuration to fail the build if warnings were found.
- Added a suppression list for checkstyle.
- Use Google's ruleset for checkstyle.


pom.xml
```diff
@@ -81,24 +83,25 @@
...
           <linkXRef>false</linkXRef>
+          <suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
+          <suppressionsFileExpression>checkstyle.suppressions.file</suppressionsFileExpression>
+          <includeTestSourceDirectory>true</includeTestSourceDirectory>
         </configuration>
         <executions>
           <execution>
             <id>validate</id>
-            <phase>validate</phase>
+            <phase>package</phase>
```
Sachpat: Just curious, what does this change do?

artem-smotrakov: The `validate` phase goes before the `package` stage. In particular, the `validation` phase is run before the tests run. In means that `checkstyle` runs before tests, and if there are any style issues, then the build stops before running tests. Now imagine that you develop a new feature and want to run tests quickly. If you introduced some style issues, you'll have to fix them in order to run your tests. That may be annoying. If checkstyle runs in the `package` phase, then the tests run before. That's why I updated the phase.

Sachpat: Great, thanks for the explanation!

src/main/java/com/sap/sgs/phosphor/fosstars/data/IsEclipse.java
```diff
@@ -12,20 +12,23 @@
...
+   * @param where An organization or username on GitHub.
    */
   public IsEclipse(String where) {
     this.where = where;
   }
 
   @Override
   public IsEclipse update(ValueSet values) {
-    System.out.println("[+] Figuring out if the project belongs to the Eclipse Software Foundation ...");
+    System.out.println(
```
Sachpat: You want these sysouts?

artem-smotrakov: Yes, this output is used in the demo tool. It's better to use a logger but it's out of scope here.

Sachpat: https://github.com/SAP/fosstars-rating-core/issues/35 yes I did.

artem-smotrakov:

Issue #37, opened 2020-03-02
RatingRepository should not use the deprecated ObjectMapper.enableDefaultTyping() method


artem-smotrakov:
The `ObjectMapper.enableDefaultTyping()` method has been deprecated, but the `RatingRepository` class still uses it:

```java
  /**
   * An ObjectMapper for serialization and deserialization ratings to JSON.
   */
  private static final ObjectMapper MAPPER = new ObjectMapper();

  static {
    MAPPER.enableDefaultTyping();
  }
```

Instead, the `RatingRepository` class should use the new version of the `enableDefaultTyping()` method which allows specifying a validator.

PR #38, opened 2020-03-03
Load a tuned OssSecurityScore from a file

### LINKED ISSUE #36: Load OssSecurityScore from a file
Test vectors for scores have been recently introduced. Now we can tune scores separately from ratings. That gives more flexibility.

The `RatingRepository` loads an instance of `OssSecurityRating` from a file including `OssSecurityScore`. The `OssScores` also provides an instance of `OssSecurityRating` which has all weights set to 1.0. That's not correct. Instead, the `RatingRepository` should load scores from files. Then, the `RatingRepository` should offer the loaded scores to the client code.

Here is what needs to be done:

- Update the `RatingRepository` class to load scores.
- Store `OssSecurityScore` to a file.
- Remove `OssScores` class.

artem-smotrakov:

- Stored a tuned instance of `OssSecurityScore` to a file.
- Updated `RatingRepository` to load a tuned instance of `OssSecurityScore` from a file.
- Updated `OssSecurityScoreTestVectors.ipynb` to tune the open-source security score
- Renamed `get()` to `rating()` in `RatingRepository`.
- Removed redundant test vectors for `OssSecurityRating`.
- Removed the `OssScore` class.
- Updated tests.

This closes #36


src/main/java/com/sap/sgs/phosphor/fosstars/model/RatingRepository.java
```diff
@@ -20,6 +23,10 @@
  */
 public class RatingRepository {
 
+  private interface RatingFactory {
+    Rating create() throws IOException;
+  }
+
   /**
    * A logger.
    */
```
sourabhsparkala: Can we replace MAPPER.enableDefaultTyping() (Depracated) to  MAPPER.activateDefaultTyping(MAPPER.getPolymorphicTypeValidator());?

artem-smotrakov: Yeah, that's a good point. I agree. Yesterday I created #37 to address that. I would prefer to fix it separately.

artem-smotrakov: I am not sure it `MAPPER.activateDefaultTyping(MAPPER.getPolymorphicTypeValidator());` is enough. Maybe we'll need to configure the validator more carefully.

sourabhsparkala: 👍

Issue #34, opened 2020-02-21
It seems to be possible that there are more contributors than commits


mibo:
It seems to be possible that there are more contributors than commits (in the last three months).
In the sample project https://github.com/jdesboeufs/connect-mongo we encountered above situation

- "name":"Number of commits last three months","n":2
- "name":"Number of contributors last three months","n":3

Which leads to a `java.lang.IllegalArgumentException: More commits than contributors! How is that possible?`

Must the check [in the ProjectActivityScore](https://github.com/SAP/fosstars-rating-core/blob/d7ec6024537cccb13c22f221968231a527efd898/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/ProjectActivityScore.java#L139) changed/removed?

artem-smotrakov:
Yeah, looks like this check is based on the wrong assumption:

```java
if (m > n) {
      throw new IllegalArgumentException("More commits than contributors! How is that possible?");
    }
```

Let's remove the check.

PR #45, opened 2020-03-04
Number of commits can be greater than number of contributors

sourabhsparkala:
Fixes https://github.com/SAP/fosstars-rating-core/issues/34

### DIFF
--- a/src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/ProjectActivityScoreTest.java
+++ b/src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/ProjectActivityScoreTest.java
@@ -109,9 +109,14 @@ public class ProjectActivityScoreTest {
     PROJECT_ACTIVITY.calculate(values(0, 1));
   }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/ProjectActivityScoreTest.java
```diff
@@ -104,15 +104,20 @@ public void commitsWithoutContributors() {
     PROJECT_ACTIVITY.calculate(values(1, 0));
   }
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test
```
artem-smotrakov: I think that this case is still invalid. The score should throw an exception here. Or, can you come up with an example which proves the opposite?

sourabhsparkala: This might be a special case, but you are right. Contributor without any commit also doesn't add to our usecase.

artem-smotrakov:
Thanks for fixing this @sourabhsparkala !

PR #46, opened 2020-03-04
Updated groupId and artifactId

### LINKED ISSUE #44: Rename the artifactId and reset version
Let's update the artifact ID to `fosstars-rating-core` and reset the version to `0.1.0`.

artem-smotrakov:
Updated `pom.xml` file:

- Updated groupId to `com.sap.phosphor`.
- Use `fosstars-rating-core` as artifactId.
- Reset the version to `0.1.0`.

Fixes #44

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,9 +4,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>com.sap.sgs.phosphor.fosstars</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -5,8 +5,8 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.sap.sgs.phosphor.fosstars</groupId>
-  <artifactId>fosstars-model</artifactId>
```
Sachpat: As discussed with @mibo , how about haveing the groupId com.sap.phosphor.

artem-smotrakov: Yup.

mibo: 👍

PR #47, opened 2020-03-05
RatingRepository should return only immutable ratings

### LINKED ISSUE #40: RatingRepository should return immutable ratings
At the moment, only the `MutableWeight` class implements the `Weight` interface. Therefore, all ratings, which are returned by the `RatingRepository`, use this implementation. The `MutableWeight` is a mutable class which makes it possible to modify the ratings returned by the `RatingRepository`. The ratings may be modified either intentionally or accidentally. To prevent this from happening. the `RatingReposotory` class should return only immutable ratings.

artem-smotrakov:

- Introduced new `isImmutable()` and `makeImmutable()` methods in the `Parameter` and `Tunable` interfaces.
- Introduced `MakeImmutable` and `ImmutabilityChecker` visitors.
- Updated `RatingRepository` to return only immutable ratings.
- Updated the existing tests and added new ones.

This fixes #40


src/main/java/com/sap/sgs/phosphor/fosstars/model/other/ImmutabilityChecker.java
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Check if an object is immutable.
+   *
+   * @param object The object to be checked.
+   */
+  private void checkImmutability(Object object) {
+    if (object instanceof Tunable) {
+      Tunable tunable = (Tunable) object;
+      if (!tunable.isImmutable()) {
+        isImmutable = false;
```
Sachpat: may be we can write it as

isImmutable  =  tunable.isImmutable();

?

artem-smotrakov: Unfortunately no. The flag must not be dropped if `isImmutable()` ever returned false.

artem-smotrakov:

Issue #48, opened 2020-03-05
Transparent rating calculation


artem-smotrakov:
Currently a rating just returns a score and a label. But it doesn't explain how the score value was actually calculated. It would be good if ratings could explain how they calculate scores.

Here is several update which may be considered:

- Update `ScoreValue` to store feature values and sub-score values.
- Add a description to `ScoreValue`.
- Update scores and ratings to provide info how they calculate values.

PR #52, opened 2020-03-06
Updated ScoreValue to store a list of used values

artem-smotrakov:

- Updated the `ScoreValue` to store a set of used values.
- Updated the existing scores to put used values to score values.
- Updated the existing tests and added new ones.

This is a part of #48

### DIFF
--- a/src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OssSecurityScoreTest.java
+++ b/src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OssSecurityScoreTest.java
@@ -1,11 +1,34 @@
 package com.sap.sgs.phosphor.fosstars.model.score.oss;
 
+import static com.sap.sgs.phosphor.fosstars.TestUtils.assertScore;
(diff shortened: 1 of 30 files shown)


src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OssSecurityScoreTest.java
```diff
@@ -21,4 +44,31 @@ public void serializeAndDeserialize() throws IOException {
     assertEquals(score, clone);
```
sourabhsparkala: Do we need a test to verify if the Used Values are properly de-/serializing?

artem-smotrakov: Yeah, we have `ScoreValueTest.serializeAndDeserialize()`.

sourabhsparkala: 👍

PR #55, opened 2020-03-10
Fixed Confidence.make() and introduced PrettyPrinter

### LINKED ISSUE #53: Confidence.make() makes a wrong confidence
If only unknown values are passed to the `Confidence.make()` method, then it returns `Confidence.MAX` but should return `Confidence.MIN`. In other words, it produces inverted confidence values.

artem-smotrakov:

- Fixed #53 
- Updated the `WeightedCompositeScore` class to use sub-score values as used values in a `ScoreValue` instance which it returns.
- Added a new `PrettyPrinter` class which makes output of `SecurityRatingCalculator` more verbose.
- Added tests.

This is a part of #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,82 +46,114 @@ mvn clean install
 
 There is a command line tool which takes a URL to a project on GitHub, fetches data about it,
 and calculates a security rating.
-The tool can interact with a user to get more data about the project.
 
 The tool can be run with a command like the following:
 
 ```
-java -jar target/fosstars-github-rating-calc.jar --token ${TOKEN} --url https://github.com/apache/poi
+java -jar target/fosstars-github-rating-calc.jar --token ${TOKEN} --url https://github.com/apache/poi --no-questions
 ```
 
 The `TOKEN` variable contains a token for accessing the GitHub API.
 You can create a personal token in the
 [settings/tokens](https://github.com/settings/tokens) tab in your profile on GitHub.
 
-The tool is a bit interactive, and can ask several question. The dialog looks like the following:
+The output is going to look like the following:
 
 ```
 [+] Project: https://github.com/apache/poi
-[+] Let's get info about the project and calculate a rating
-[?] Are you aware about any unpatched vulnerability in the project? (yes/no)
->>> no
-Downloading files at Wed Oct 23 13:35:52 CEST 2019
-Using local NVD cache as last update was within two hours
+[+] Let's get info about the project and calculate a security rating
+[+] Counting how many commits have been done in the last three months ...
+[+] Counting how many people contributed to the project in the last three months ...
+[+] Counting how many stars the project has ...
+[+] Counting how many watchers the project has ...
+[+] Figuring out when the first commit was done ...
+[+] Figuring out when the project started ...
+[+] Figuring out if the project has a security team ...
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -62,28 +62,96 @@ The tool is a bit interactive, and can ask several question. The dialog looks li
 
 ```
 [+] Project: https://github.com/apache/poi
-[+] Let's get info about the project and calculate a rating
+[+] Let's get info about the project and calculate a security rating
+[+] Counting how many commits have been done last three months ...
```
sourabhsparkala: Counting how many commits have been done in last three months ...

README.md
```diff
@@ -62,28 +62,96 @@ The tool is a bit interactive, and can ask several question. The dialog looks li
...
+[+] Counting how many people contributed to the project last three months ...
```
sourabhsparkala: Counting how many people contributed to the project in last three months ...

README.md
```diff
@@ -100,26 +168,94 @@ The output looks like the following:
 
 ```
 [+] Project: https://github.com/apache/poi
-[+] Let's get info about the project and calculate a rating
-Downloading files at Wed Oct 23 13:41:21 CEST 2019
-Using local NVD cache as last update was within two hours
+[+] Let's get info about the project and calculate a security rating
+[+] Counting how many commits have been done last three months ...
```
sourabhsparkala: Counting how many commits have been done in last three months ...

README.md
```diff
@@ -100,26 +168,94 @@ The output looks like the following:
...
+[+] Counting how many people contributed to the project last three months ...
```
sourabhsparkala: Counting how many people contributed to the project in last three months ...

src/main/java/com/sap/sgs/phosphor/fosstars/model/Confidence.java
```diff
@@ -85,7 +85,7 @@ static double make(Value... values) {
       return MAX;
     }
 
-    return (MAX * unknown) / values.length;
+    return MAX * (values.length - unknown) / values.length;
```
sourabhsparkala: Is there a test case to check
when values.length = 0?

Test case for this function make()

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/ScoreValue.java
```diff
@@ -80,12 +80,16 @@ public ScoreValue(
     this.usedValues = new ArrayList<>(Objects.requireNonNull(usedValues, "Values can't be null!"));
   }
 
-  @Override
   @JsonGetter("score")
-  public Score feature() {
+  public Score score() {
```
sourabhsparkala: Would this be more appropriate if the method name is value()?

README.md
artem-smotrakov: I am not sure whether the current version is correct since I am not a native speaker. I'll update the sentences if you think they should use "in".

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/ScoreValue.java
artem-smotrakov: This method doesn't return the value. It returns a score which calculated the value.

PR #57, opened 2020-03-12
Renamed SecurityLabel.OKAY to SecurityLabel.MODERATE

### LINKED ISSUE #56: Rename SecurityLabel.OKAY to SecurityLabel.MODERATE
The difference between GOOD and OKAY looks not that clear (in `com.sap.sgs.phosphor.fosstars.model.rating.oss.OssSecurityRating` the `SecurityLabel` enum).

I suggest to rename the `OKAY` to `MODERATE`. 
IMHO this fits better between `GOOD` and `BAD`.
Result would be:
```
public enum SecurityLabel implements Label {
  BAD, MODERATE, GOOD
}
```

See related PR #57 which I have created.

mibo:
Fixes #56

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
@@ -15,7 +15,7 @@ import com.sap.sgs.phosphor.fosstars.model.score.oss.OssSecurityScore;
 public class OssSecurityRating extends AbstractRating {
 
   public enum SecurityLabel implements Label {
-    BAD, OKAY, GOOD
+    BAD, MODERATE, GOOD
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mibo:
Rebased and changed commit message to follow CONTRIBUTING guidelines.

artem-smotrakov:
Looks good to me.

mibo:
> Thanks for addressing this @mibo . I was wondering if any docs need to be updated as well. Looks like the docs don't mention the old `OKAY` label. Only the README file contains a sample output from the demo tool which mentions the old label. I am going to update the sample output anyway, so no need to update the README now.

Perhaps in the future a section can be added describing the labels for the available ratings.
Or the enum itself could/should contain some JavaDoc or a description field. 
But both are out of scope for this PR  🙂

artem-smotrakov:
@mibo Yes, the docs for the open-source security rating doesn't currently provide too much info. I agree that more details should be added. The javadoc would not make it worse either.

PR #59, opened 2020-03-13
Added descriptions for scores and weights to score values

artem-smotrakov:
Here is the main updates:

- Added a new `Score.description()` method.
- Added descriptions for several scores.
- The `ScoreValue` class now stores a weight for a sub-score.
- Updated the `PrettyPrinter` class to print out the introduced info.

Note that not all existing scores have been updated with descriptions. I am planning to update the scores soon, so I think it's better to think about their descriptions after that.

This changeset partly addresses #48

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
@@ -1,0 +1,98 @@
+package com.sap.sgs.phosphor.fosstars.model.math;
+
+import java.util.function.Function;
+import org.apache.commons.math3.analysis.function.Logistic;
+
+/**
+ * This is a helper class math.
+ */
+public class MathHelper {
+
+  /**
+   * Invert a value of a logistic function on a specified interval.
+   *
+   * @param func The logistic function to be inverted.
+   * @param a The beginning of the interval.
+   * @param b The end of the interval.
+   * @param value The value to be inverted.
+   * @param precision The precision for comparison of doubles.
+   * @return The argument of the function, which belongs to the specified interval [a, b],
+   *         which produces the closest value to the specified value.
+   * @throws IllegalArgumentException If the parameters are incorrect.
+   */
+  public static int invert(Logistic func, int a, int b, double value, double precision) {
+    if (a > b) {
+      throw new IllegalArgumentException("a > b");
+    }
+    double funcA = func.value(a);
+    if (equals(funcA, value, precision)) {
+      return a;
+    }
+    if (a == b) {
+      throw new IllegalArgumentException(String.format("a == b and func(a) != %2.2f", value));
+    }
+    if (Double.compare(funcA, value) > 0) {
+      throw new IllegalArgumentException(String.format(
+          "func(%d) is %2.2f which is greater than %2.2f", a, funcA, value));
+    }
+    double funcB = func.value(b);
+    if (equals(funcB, value, precision)) {
+      return b;
+    }
+    if (Double.compare(funcB, value) < 0) {
+      throw new IllegalArgumentException(String.format(
+          "func(%d) is %2.2f which is less than %2.2f", b, funcB, value));
+    }
+
+    return binarySearch(func::value, a, b, value, precision);
+  }
+
+  /**
+   * Inverses a monotonic function on a specified interval using binary search.
+   *
+   * @param func The function to be inverted.
+   * @param a The beginning of the interval.
+   * @param b The end of the interval.
+   * @param value The value of the function to be inverted.
+   * @param precision The precision for comparison of doubles.
+   * @return The argument of the function, which belongs to the specified interval [a, b],
+   *         which produces the closest value to the specified value.
+   */
+  private static int binarySearch(
+      Function<Integer, Double> func, int a, int b, double value, double precision) {
+
+    while (b - a > 1) {
+      int m = a + (Math.abs(b - a)) / 2;
+      double funcM = func.apply(m);
+      if (equals(funcM, value, precision)) {
+        return m;
+      }
+      if (Double.compare(value, funcM) > 0) {
+        a = m;
+      } else {
+        b = m;
+      }
+    }
+
+    double funcA = func.apply(a);
+    double funcB = func.apply(b);
+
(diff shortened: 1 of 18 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
```diff
@@ -0,0 +1,122 @@
...
+   *         which produces the closest value to the specified value.
+   * @throws IllegalArgumentException If the parameters are incorrect.
+   */
+  public static int invert(
+      Function<Integer, Double> func, int a, int b, double value, double precision) {
+
+    if (a > b) {
+      throw new IllegalArgumentException("a > b");
+    }
+    double funcA = func.apply(a);
```
Sachpat: why don't you name it differently as valueA or something? because funcA is confusing if it's a function or a double value?

src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (Double.compare(funcA, value) > 0) {
+      throw new IllegalArgumentException(String.format(
+          "func(%d) is %2.2f which is greater than %2.2f", a, funcA, value));
+    }
+    double funcB = func.apply(b);
+    if (equals(funcB, value, precision)) {
+      return b;
+    }
+    if (Double.compare(funcB, value) < 0) {
```
Sachpat: same here, it makes it bit confusing

src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
artem-smotrakov: `funcA` is supposed to mean "func(a)" which means "a value of the function in the point a". That's why it is `funcA`. For me it's not confusing. If you think it is confusing, please suggest how exactly you'd like to name it (with the meaning above). The shorter the better.

artem-smotrakov: `valueA` doesn't look better to me.

Sachpat: It's of-course not confusing to you because you are the author. :) . Please see the below line:
if (equals(funcB, value, precision)) {

this looks like you are comparing a function to a value. 

Why don't you name it as 

functionValueA?

Sachpat: As I said earlier, if you are writing the code for yourself, fine for me. If you are writing the code for others to contribute or other to understand also, consider changing it.

artem-smotrakov: I am not saying that the current name it not confusing. I am saying it is not confusing to me (for whatever reason). That's why I asked what you think would be a better name. Preferably a short name. How about `funcOfA` instead of `functionValueA`?

Sachpat: Hmm, isn't it a Value after applying function to a Value A? or it's still a function?

Sachpat: To my understanding, it's an inverted Value of A right?

artem-smotrakov: It is a value but I think `funcOfA` is clear enough. If it is still not, then I suggest `funcValueA`. Does it work for you?

Sachpat: No funOfA is fine :) Thanks for the clarification.

artem-smotrakov: Taking into account your points, I will update it to `funcValueA` nevertheless.

Sachpat: No, No, now for me funcOfA makes more sense.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/PrettyPrinter.java
```diff
@@ -16,24 +29,84 @@
...
     StringBuilder sb = new StringBuilder();
-    sb.append(String.format("[+] %sScore:........%s%n", indent, scoreValue.score().name()));
+
+    if (!isMainScore) {
+      sb.append(String.format("[+] %sSub-score:....%s%n", indent, nameOf(scoreValue.score())));
+    } else {
+      sb.append(String.format("[+] %sScore:........%s%n", indent, nameOf(scoreValue.score())));
+    }
+
+    if (!scoreValue.score().description().isEmpty()) {
```
Sachpat: Is there a possiblity of scoreValue or  scoreValue.score() being null here?

src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
artem-smotrakov: Sure, no problem. I have updated the variable name to `funcOfA`, `funcOfB` and so on. Please let me know if I missed anything.

src/main/java/com/sap/sgs/phosphor/fosstars/model/math/MathHelper.java
artem-smotrakov: Yup, addressed.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/PrettyPrinter.java
artem-smotrakov: Fortunately, no. The `scoreValue.score()` can't return null because the constructor of the `ScoreValue` checks the score for null. `scoreValue` can't be null either because it comes from the `RatingValue` which also checks its score value for null in the constructor.

artem-smotrakov:
> Thanks for addressing the comments Artem, looks good to me.

PR #61, opened 2020-03-13
Added explanations to ScoreValue

### LINKED ISSUE #48: Transparent rating calculation

### LINKED ISSUE #60: Add Value.processIfUnknown() method for processing unknown values
The `Value` interface already defines a convenient `processIfKnown()` method which calls a processor if the value is known. It would be good to adds a new `processIfUnknown()` which calls a processor if the value is unknown. The new method would help to simplify scores.

artem-smotrakov:

- Added a new `ScoreValue.explanation()` method which returns a list of notes which explain how the score value was calculated.
- Added a new `ScoreValue.explain()` method which adds an explanation note for the score value.
- Updated a couple of scores to explain how it calculates scores.
- Updated the `PrettyPrinter` to print out explanations.
- Added a new `Value.processIfUnknown()` method.
- Replaced "last three months" with "in the last three months" in all feature names (it caused updates to the notebooks and test vectors).
- Updated the sample output of the demo tool in README.
- Updated tests.

Please note that not all scores explain how they calculate values. I am planning to update some scores in the near future. After that, we may want to make them explain themselves.

This closes #48
This also closes #60


src/main/java/com/sap/sgs/phosphor/fosstars/model/Value.java
```diff
@@ -51,13 +51,21 @@
   T get();
 
   /**
-   * Call a processor to process the value if the value is known.
+   * Call a processor to process the value if it's known.
    *
    * @param processor The processor to be called.
    * @return This Value instance.
    */
   Value<T> processIfKnown(Processor<T> processor);
```
mibo: Not directly related to this commit but the `Processor` interface could have the  `@FunctionalInterface` annotation.

artem-smotrakov: Okay, let's add it.

artem-smotrakov:
> What I have seen in this PR looks good 

Thanks for the review @mibo 
 
> Only things I found are not directly related to this PR 
> Like the `@FunctionalInterface` and that in the `com.sap.sgs.phosphor.fosstars.tool.github.SecurityRatingCalculator#USAGE` must be updated.
> 

I'll update it.

> Furthermore I had the feeling that the `SecurityRatingCalculator` part deserves an own maven (sub)module.
> IMHO this would make it clear what is part of the `rating-core` library and what is build on top of the library (use it as dependency) as e.g. the`SecurityRatingCalculator`.
> Perhaps I should create an issue to discuss if other have a similar opinion (or not).

Okay, let's think about it.

PR #64, opened 2020-03-17
Introduced DependencyScanScore

### LINKED ISSUE #63: Add DependencyScanScore
Currently, the `ProjectSecurityTestingScore` is a feature-based score. It uses the `SECURITY_REVIEWS_DONE` and `SCANS_FOR_VULNERABLE_DEPENDENCIES` features. However, it needs to be updated to use the `LgtmScore`, see #24. Therefore, it would be better to make the `ProjectSecurityTestingScore` to use only scores. In order to implement that, the `SECURITY_REVIEWS_DONE` is going to be removed from the score, see #25. Second, the `SCANS_FOR_VULNERABLE_DEPENDENCIES` feature needs to be wrapped in to a new `DependencyScanScore`.

artem-smotrakov:
Here is a list of updates:

- Added a new `DependencyScanScore` which is based on `SCANS_FOR_VULNERABLE_DEPENDENCIES` feature.
- Added a verification procedure for the new score.
- Added a new `DependencyScanScoreTest`.

The `DependencyScanScore` is necessary to make the `ProjectSecurityTestingScore` use only sub-scores in near future.

This fixes #63


artem-smotrakov:

PR #65, opened 2020-03-17
Removed SECURITY_REVIEWS_DONE feature from OssSecurityRating

### LINKED ISSUE #25: Remove SECURITY_REVIEWS_DONE feature from OssSecurityRating
It's hard to gather data for the `SECURITY_REVIEWS_DONE` in an automated way. Currently we maintain a list of security reviews in a JSON file.

Once we have more features, which contribute to the `ProjectSecurityTestingScore` (see for example #24), we can remove the `SECURITY_REVIEWS_DONE` from the `OssSecurityRating` until we get enough data to fill it out.

artem-smotrakov:
This patch removes the `SECURITY_REVIEWS_DONE` feature from the `OssSecurityScore` because it's hard to gather data about security reviews. However, the feature still stays in the `OssFeatures`.

The `OssSecurityRating` will be soon updated to use the new `LgtmScore` and `DependencyScanScore`.

This fixes #25


artem-smotrakov:
> Overall looks good.
> 
> Just observed that SECURITY_REVIEWS_DONE still seems to be used in some
> 
> * Tests like PrettyPrinterTest, OssSecurityScoreTuningWithCMAESTest
> * Classes like PrettyPrinter
> * Also please check README.doc, if it is necessary to remove it there.

Yup. Good point. Although it's not a goal to remove the feature completely, I'll check which classes should be updated.

artem-smotrakov:
@sourabhsparkala I have updated other classes which use the feature. Now only `SecurityReviewsDoneValue` and `PrettyPrinter` refer to the feature. I would not update these classes. Please have a look at the updated changeset.

PR #58, opened 2020-03-13
Take into account signed commits in the open-source security rating

sourabhsparkala:
- Introduce a new boolean feature which tell whether a project uses signed commits.
- Include the new feature to the security awareness score unless there is a better place.
- Update the test vectors for the updated score with the new feature. Re-tune the scores if necessary.
- Update the test vectors for the open-source security score and rating to use the new feature.
- Update one of the data providers to fill out the new feature in the demo tool. Or, create a new one if necessary. Make sure that the demo tool works fine.
- Add new tests.

PR #66, opened 2020-03-17
Implement a DataProvider to generate feature USES_VERIFIED_SIGNED_COMMITS

### LINKED ISSUE #54: Take into account signed commits in the open-source security rating
GitHub allows signing commits to make sure that the changes come from a trusted store:

- [Managing commit signature verification](https://help.github.com/en/github/authenticating-to-github/managing-commit-signature-verification)
- [Commits API](https://developer.github.com/v3/repos/commits/)

It may be good to update the open-source security rating to check whether commits are signed or not.

Here is a list of things to be done:

- Introduce a new boolean feature which tell whether a project uses signed commits.
- Include the new feature to the security awareness score unless there is a better place.
- Update the test vectors for the updated score with the new feature. Re-tune the scores if necessary.
- Update the test vectors for the open-source security rating to use the new feature.
- Update one of the data providers to fill out the new feature in the demo tool. Or, create a new one if necessary. Make sure that the demo tool works fine.
- Add new tests.

sourabhsparkala:
This PR is the next step after #58 

- Implementing a DataProvider which would collect the list of commits (within in a year) for a oss-project.
- Compute a density ratio `Total number of verified signed commits / Total number of commits within an year`.
- If the density ratio crosses a certain Percentage Threshold then:
  - USES_VERIFIED_SIGNED_COMMITS = `True`
  - else USES_VERIFIED_SIGNED_COMMITS = `False`

This closes #54

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
@@ -1,0 +1,231 @@
+package com.sap.sgs.phosphor.fosstars.data.github;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_VERIFIED_SIGNED_COMMITS;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sap.sgs.phosphor.fosstars.model.Value;
+import com.sap.sgs.phosphor.fosstars.model.ValueSet;
+import com.sap.sgs.phosphor.fosstars.model.value.BooleanValue;
+import com.sap.sgs.phosphor.fosstars.model.value.UnknownValue;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.Optional;
+import org.apache.http.Header;
+import org.apache.http.HttpHeaders;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.entity.ContentType;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.kohsuke.github.GitHub;
+
+/**
+ * This data provider checks if a project uses verified signed commits. The check would be based on
+ * a density ratio compared against a percentage threshold
+ * ({@link #SIGNED_COMMIT_PERCENTAGE_THRESHOLD}).
+ */
+public class UsesSignedCommits extends AbstractGitHubDataProvider {
+
+  /**
+   * For parsing JSON.
+   */
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  /**
+   * 1 year in millis.
+   */
+  private static final long DELTA = 365 * 24 * 60 * 60 * 1000L;
+
+  /**
+   * The threshold in % of number of verified signed commits against total number of commits in a
+   * year.
+   */
+  private static final double SIGNED_COMMIT_PERCENTAGE_THRESHOLD = 90;
+
+  /**
+   * Maximum number of items to be populated per page during the call to GitHub API.
+   */
+  private static final int MAX_ITEMS_PER_PAGE = 100;
+
+  /**
+   * First page number during the call to GitHub API.
+   */
+  private static final int FIRST_PAGE = 1;
+
+  /**
+   * A GitHub token provided by user to access the GitHub API.
+   */
+  private final String githubToken;
+
+  /**
+   * The date format to be used during GitHub API call.
+   */
+  private static final SimpleDateFormat DATE_FORMAT =
+      new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US);
+
+  /**
+   * Initializes a data provider.
+   *
+   * @param where A GitHub organization or user name.
+   * @param name A name of a repository.
+   * @param github An interface to the GitHub API.
+   * @param githubToken provided by user to access GitHub API.
+   */
+  public UsesSignedCommits(String where, String name, GitHub github, String githubToken) {
+    super(where, name, github);
+    this.githubToken = githubToken;
+  }
+
+  @Override
+  public UsesSignedCommits update(ValueSet values) throws IOException {
+    Objects.requireNonNull(values, "Hey! Values can't be null!");
+    System.out.println("[+] Figuring out if the project uses verified signed commits ...");
+
+    Optional<Value> something = cache().get(url, USES_VERIFIED_SIGNED_COMMITS);
+    if (something.isPresent()) {
+      values.update(something.get());
+      return this;
+    }
+
+    Value<Boolean> signedCommits = usesSignedCommits();
+    values.update(signedCommits);
+    cache().put(url, signedCommits, tomorrow());
+
+    return this;
+  }
+
+  /**
+   * It gathers the list of commits within one year from the current date. Then it calculates
+   * density ratio. The density ratio is the percentage calculation of number of verified signed
+   * commits against the total number of commits in the year. If the density ratio in % crosses a
+   * certain percentage threshold ({@link #SIGNED_COMMIT_PERCENTAGE_THRESHOLD}), then it says that
+   * the project uses verified signed commits. Otherwise, it says that the project does not use
+   * verified signed commits.
+   * 
+   * @return {@link BooleanValue} value of the feature {@link #USES_VERIFIED_SIGNED_COMMITS}.
+   */
+  private Value<Boolean> usesSignedCommits() {
+    try {
+      return new BooleanValue(USES_VERIFIED_SIGNED_COMMITS, askGithub());
+    } catch (IOException e) {
+      System.out.printf("[x] Couldn't fetch data from api.github.com!%n");
+      System.out.printf("[x]     %s%n", e);
+      return UnknownValue.of(USES_VERIFIED_SIGNED_COMMITS);
+    }
+  }
+
+  /**
+   * Check if the project uses signed verified commits.
+   * 
+   * @return boolean True if the project uses verified signed commits. Otherwise, False
+   * @throws IOException #{@link HttpClient} may throw an exception during REST call.
+   * @see: GitHub Issue: <a link="https://github.com/github-api/github-api/issues/737">Get commit or
+   *       tag signature verified flag</a> and the associated
+   *       <a link= "https://github.com/github-api/github-api/pull/738">Pull Request</a>.
+   */  
+  private boolean askGithub() throws IOException {
+    // TODO: The data gathering process here from GitHub API will have to be merged with other
+    // related data providers. More information can be found in
+    // https://github.com/SAP/fosstars-rating-core/issues/69
+    
+    int signedCounter = 0;
+    int counter = 0;
+    String commitDate = DATE_FORMAT.format(new Date(System.currentTimeMillis() - DELTA));
+    int pageNumber = FIRST_PAGE;
+    Page page;
+    
+    do {
+      page = nextPage(commitDate, pageNumber++);
+      
+      for (JsonNode commitNode : page.commitNodes) {
+        counter++;
+        JsonNode verifiedNode = commitNode.findPath("verified");
+        if (verifiedNode != null && verifiedNode.booleanValue()) {
+          signedCounter++;
+        }
+      }
+    } while (page.hasNextPage);
+    
+    return counter > 0 && signedCounter > 0
+        && (signedCounter * 100 / counter) >= SIGNED_COMMIT_PERCENTAGE_THRESHOLD;
+  }
+
+  /**
+   * Does a REST API call to public URL <a link="https://api.github.com">GitHub API</a> to list all
+   * the commits after a specific date.
+   * 
+   * @param commitDate Only commits after this date will be returned. 
+   *                   This is a timestamp in ISO 8601, format: YYYY-MM-DDTHH:MM:SSZ
+   * @param pageNumber Index of the page.
+   * @return The {@link Page} object.
+   * @throws IOException #{@link HttpClient} may throw an exception during REST call.
+   */
+  private Page nextPage(String commitDate, int pageNumber) throws IOException {
+    try (CloseableHttpClient client = httpClient()) {
+      HttpGet httpGetRequest = buildRequest(commitDate, pageNumber);
+      try (CloseableHttpResponse response = client.execute(httpGetRequest)) {
+        return new Page(hasNextPage(response), MAPPER.readTree(response.getEntity().getContent()));
+      }
+    }
+  }
+
+  /**
+   * Builds a {@link HttpGet} request.
+   * 
+   * @param commitDate This is a timestamp in ISO 8601, format: YYYY-MM-DDTHH:MM:SSZ
+   * @param pageNumber Index of the page.
+   * @return The {@link HttpGet} object.
+   */
+  private HttpGet buildRequest(String commitDate, int pageNumber) {
+    String url =
+        String.format("https://api.github.com/repos/%s/%s/commits?since=%s&page=%s&per_page=%s",
+            where, name, commitDate, pageNumber, MAX_ITEMS_PER_PAGE);
+
+    HttpGet httpGetRequest = new HttpGet(url);
+    httpGetRequest.addHeader(HttpHeaders.ACCEPT, ContentType.APPLICATION_JSON.getMimeType());
+
+    if (githubToken != null && !githubToken.isEmpty()) {
+      httpGetRequest.addHeader(HttpHeaders.AUTHORIZATION, String.format("token %s", githubToken));
+    }
+
+    return httpGetRequest;
+  }
+
+  /**
+   * Returns an HTTP client.
+   */
+  CloseableHttpClient httpClient() {
+    return HttpClients.createDefault();
+  }
+
+  /**
+   * Check if the next page is available.
+   * 
+   * @param response {@link CloseableHttpResponse} from the GET Request.
+   * @return True if the next page is available, false otherwise.
+   */
+  private boolean hasNextPage(CloseableHttpResponse response) {
+    Header header = response.getFirstHeader("link");
+    return header != null && header.toString().contains("rel=\"next\"");
+  }
+
+  /**
+   * This class is used to store the commit nodes and hasNextPage indication after each call to
+   * GitHub API. After the call the variables hasNextPage and commitNodes gets populated.
+   */
+  private static class Page {
+
+    private final boolean hasNextPage;
+    private final JsonNode commitNodes;
+
+    private Page(boolean hasNextPage, JsonNode commitNodes) {
+      this.hasNextPage = hasNextPage;
+      this.commitNodes = commitNodes;
+    }
+  }
+}
--- a/src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
+++ b/src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
@@ -1,0 +1,116 @@
+package com.sap.sgs.phosphor.fosstars.data.github;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_VERIFIED_SIGNED_COMMITS;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+
+import com.sap.sgs.phosphor.fosstars.data.ValueCache;
+import com.sap.sgs.phosphor.fosstars.model.value.ValueHashSet;
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.http.HttpEntity;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.Test;
+import org.kohsuke.github.GitHub;
+
+public class UsesSignedCommitTest {
+
+  @Test
+  public void testUsesSignedCommitsTrue() throws IOException {
+    GitHub github = mock(GitHub.class);
+    UsesSignedCommits provider =
+        new UsesSignedCommits("spring-projects", "spring-integration", github,
+            null);
+    provider = spy(provider);
+    when(provider.cache()).thenReturn(new ValueCache());
+
+    CloseableHttpClient httpClient = mock(CloseableHttpClient.class);
+    when(provider.httpClient()).thenReturn(httpClient);
+
+    CloseableHttpResponse response = mock(CloseableHttpResponse.class);
+    when(httpClient.execute(any())).thenReturn(response);
+
+    HttpEntity entity = mock(HttpEntity.class);
+    when(response.getEntity()).thenReturn(entity);
+
+    try (InputStream content = getClass().getResourceAsStream("UsesSignedCommitTrue.json")) {
+      when(entity.getContent()).thenReturn(content);
+
+      ValueHashSet values = new ValueHashSet();
+      assertEquals(0, values.size());
+
+      provider.update(values);
+
+      assertEquals(1, values.size());
+      assertTrue(values.has(USES_VERIFIED_SIGNED_COMMITS));
+      assertTrue(values.of(USES_VERIFIED_SIGNED_COMMITS).isPresent());
+      assertFalse(values.of(USES_VERIFIED_SIGNED_COMMITS).get().isUnknown());
+      assertEquals(USES_VERIFIED_SIGNED_COMMITS.value(true),
+          values.of(USES_VERIFIED_SIGNED_COMMITS).get());
+    }
+  }
+
+  @Test
+  public void testUsesSignedCommitsFalse() throws IOException {
+    GitHub github = mock(GitHub.class);
+    UsesSignedCommits provider =
+        new UsesSignedCommits("spring-projects", "spring-integration", github, null);
+    provider = spy(provider);
+    when(provider.cache()).thenReturn(new ValueCache());
+
+    CloseableHttpClient httpClient = mock(CloseableHttpClient.class);
+    when(provider.httpClient()).thenReturn(httpClient);
+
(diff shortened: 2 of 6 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * This data provider tries to figure out if a project uses signed commits. First, it checks if a
+ * project belongs to an organization and checks if any commits are signed or verified. It gathers
```
artem-smotrakov: I think it's better to say "... checks if a project uses verified signed commits" according to the feature description.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+ * the list of commits within one year from the current date. Then it calculates density ratio
+ * between number of signed commits vs unsigned commits. If the density ratio >= 1, then we set the
```
artem-smotrakov: I think it is better not to hardcode constants in the description. Instead, you can refer to the constants defined in the class (using javadoc markup).

Please also check the rest of javadoc.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+ * flag = true else flag = false. Foundation. Next, it tries to ask a user if {@link UserCallback}
```
artem-smotrakov: Please remove "Foundation".

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * 1 year in millis.
+   */
+  private static final long DELTA = 365 * 24 * 60 * 60 * 1000L;
+
+  /**
+   * The percentage threshold of number of signed commits against number of total commits.
+   */
+  private static final double SIGNED_COMMIT_PRECENTAGE_THRESHOLD = 0.4;
```
artem-smotrakov: I would suggest to use much higher threshold value, let's say 90% or even more. What do you think?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * A git token provided by user to access the Github API.
```
artem-smotrakov: Please change it to "A GitHub token ..." since git doesn't have tokens but GitHub does. Also, please use "GitHub" (capital H). Sorry :) Then, it would be better to rename `gitToen` to `githubToken`.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+   */
+  private final String gitToken;
+
+  /**
+   * Initializes a data provider.
+   *
+   * @param where A GitHub organization of user name.
```
artem-smotrakov: of -> or

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+  public UsesSignedCommits(String where, String name, GitHub github, String gitToken)
+      throws IOException {
+    super(where, name, github);
+    this.gitToken = gitToken;
+  }
+
+  @Override
+  public UsesSignedCommits update(ValueSet values) throws IOException {
+    Objects.requireNonNull(values, "Hey! Values can't be null!");
+    System.out.println("[+] Figuring out if the project uses any signed commits ...");
```
artem-smotrakov: Please update it to "if the project uses verified signed commits"

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Optional<Value> something = cache().get(url, USES_VERIFIED_SIGNED_COMMITS);
+    if (something.isPresent()) {
+      values.update(something.get());
+      return this;
+    }
+
+    boolean flag = densityEvaluation();
+    if (callback.canTalk()) {
```
artem-smotrakov: I don't think we need to ask a user about this feature. This is pretty clear how the data provider can fill out the feature without a user.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * It gathers the list of commits within one year from the current date. It checks if these
+   * commits are signed or not. Then it calculates density ratio between number of signed commits vs
+   * unsigned commits. If the density ratio >= 1, then we set the flag = true else flag = false.
+   * 
+   * @return boolean value indicating if the project uses signed commit.
+   * @throws IOException
+   *
```
artem-smotrakov: Please remove an extra line.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+    return this;
+  }
+
+  /**
+   * It gathers the list of commits within one year from the current date. It checks if these
+   * commits are signed or not. Then it calculates density ratio between number of signed commits vs
+   * unsigned commits. If the density ratio >= 1, then we set the flag = true else flag = false.
+   * 
+   * @return boolean value indicating if the project uses signed commit.
+   * @throws IOException
```
artem-smotrakov: Please describe the exception.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+
+    return this;
+  }
+
+  /**
+   * It gathers the list of commits within one year from the current date. It checks if these
+   * commits are signed or not. Then it calculates density ratio between number of signed commits vs
+   * unsigned commits. If the density ratio >= 1, then we set the flag = true else flag = false.
+   * 
+   * @return boolean value indicating if the project uses signed commit.
```
artem-smotrakov: Please update it to "If the project uses verified signed commits"

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+  private boolean densityEvaluation() throws IOException {
+    GHRepository repository = github.getRepository(path);
+    double signedCounter = 0;
+    double counter = 0;
+    Date date = new Date(System.currentTimeMillis() - DELTA);
+
+    for (GHCommit commit : repository.listCommits()) {
+      if (commit.getCommitDate().after(date)) {
+        counter++;
+        // TODO: Remove this commit rest call when org.kohsuke:github-api implements
```
artem-smotrakov: I would say that we may consider merging this data provider with `FirstCommit` and `ProjectStarted` data providers since all of them use commit history. That would speed up gathering data. Could you please update the TODO and open an issue for this improvement?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+        // GHCommit::verification
+        JSONObject restCommit = new JSONObject(commitInfo(commit.getSHA1()));
+        if (restCommit.getJSONObject("commit").getJSONObject("verification")
```
artem-smotrakov: May NPE occur here? Does it make sense to add null-checks?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
artem-smotrakov: If I understand correctly, `0.4` doesn't mean percentage, it means a fraction. You didn't mean 0.4%, did you? Please drop the "PRECENTAGE" from the constant name or update the constant value (by the way, there is a typo in PRECENTAGE).

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+
+    if (counter > 0 && signedCounter > 0
+        && (signedCounter / counter) >= SIGNED_COMMIT_PRECENTAGE_THRESHOLD) {
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Does a rest api call to public url https://api.github.com to gather a specific git detail
```
artem-smotrakov: Please, "REST API", "URL" (capitals).

artem-smotrakov: Please explain "to gather a specific git detail".

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+    double signedCounter = 0;
+    double counter = 0;
+    Date date = new Date(System.currentTimeMillis() - DELTA);
+
+    for (GHCommit commit : repository.listCommits()) {
+      if (commit.getCommitDate().after(date)) {
+        counter++;
+        // TODO: Remove this commit rest call when org.kohsuke:github-api implements
+        // GHCommit::verification
+        JSONObject restCommit = new JSONObject(commitInfo(commit.getSHA1()));
```
artem-smotrakov: Would it be possible to make a single request for many commits instead of sending a request for each commit?

With the current solution, it may be very likely that a user hits the rate limit.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+    return false;
+  }
+
+  /**
+   * Does a rest api call to public url https://api.github.com to gather a specific git detail
+   * 
+   * @param sha The commit SHA1 identifier which will be used as URI
+   * @return The {@link GithubCommit} object
+   * @see: <a link="https://github.com/github-api/github-api/issues/737">Issue</a>,
```
artem-smotrakov: Please replace "Issue" with the subject of the issue.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
sourabhsparkala: Good catch I have put in the NPE check

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
sourabhsparkala: You are absolutely right https://developer.github.com/v3/repos/commits/#parameters, has a `since` parameter which can be used. I will update this.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
sourabhsparkala: @artem-smotrakov , I ran a quick test on `https://github.com/spring-projects/spring-integration` as the input for the past year:

- Number of signed commits: 24
- Total number of commits: 493

Density percent : 4.8681 %. 

Looking at this I am thinking to keep the SIGNED_COMMIT_PERCENTAGE_THRESHOLD = 10. What do you think?

artem-smotrakov: @sourabhsparkala I don't think we should align the threshold value with one (or even many) projects.

The new feature answers the question "Does the project uses verified signed commits?". If I am given a project, and if I am asked this question, then I would ask myself "Are most of commits in the projects signed?". If I see that that only 4.86% of commits are signed, I would answer no to the question. Would you answer yes?

I think that "most of commits" means at least 90%. I would even say 95%. If a project require commits to be signed, then an unsigned commit should be a rare exception.

sourabhsparkala: Yes I agree. with your comment.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
artem-smotrakov: #69

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
sourabhsparkala: I have updated this. But not sure why this is still showing here.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,179 @@
...
+    double counter = 0;
+    Date date = new Date(System.currentTimeMillis() - DELTA);
+
+    // TODO: GitHub Issue: <a link= "https://github.com/SAP/fosstars-rating-core/issues/69">Unifying
+    // the data gathering process for data providers</a>
+    SimpleDateFormat simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US);
+    String commitDate = simpleDateFormat.format(date);
+
+    int pageNumber = MIN_GITHUB_PAGE;
+    while (pageNumber <= MAX_GITHUB_PAGE) {
```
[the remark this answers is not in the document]
sourabhsparkala: You are right, There seems to be no limit to the number of pages allowed. Hence I add that. But it does not make any sense to have this restriction also. https://developer.github.com/v3/guides/traversing-with-pagination/

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+   * @param date Only commits after this date will be returned. 
+   *             This is a timestamp in ISO 8601, format: YYYY-MM-DDTHH:MM:SSZ
+   * @param page Index of the page
+   * @return The {@link JsonNode} object
+   * @throws IOException #{@link HttpClient} may throw an exception during REST call.
+   * @see: GitHub Issue: <a link="https://github.com/github-api/github-api/issues/737">Get commit or
+   *       tag signature verified flag</a> and the associated
+   *       <a link= "https://github.com/github-api/github-api/pull/738">Pull Request</a>.
+   */
+  private JsonNode commitInfo(String date, int page) throws IOException {
```
artem-smotrakov: Would you mind returning `httpResponse` from this method? That would allow dropping the `hasNextPage` field (which should be at most a local variable).

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+    // TODO: Remove this method when org.kohsuke:github-api implements GHCommit::verification. More
+    // information can be found in https://github.com/github-api/github-api/issues/737
+    CloseableHttpClient client = httpClient();
```
artem-smotrakov: Please use try-with-resources to close the client.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+  /**
+   * Check if next page is available, returns true. Otherwise the last page has been reached.
+   * 
+   * @param response {@link CloseableHttpResponse} from the GET Request.
+   * @return boolean if the header "link" contains 'rel="next"'
+   */
+  boolean hasNextPage(CloseableHttpResponse response) {
+    String headerName = "link";
+    return response.getFirstHeader(headerName) != null
+        && response.getFirstHeader(headerName).toString().contains("rel=\"next\"");
```
artem-smotrakov: `headerName` should be a constant. I would even drop it and make it shorter:

```java
String header = response.getFirstHeader("link");
return header != null && header.toString().contains("rel=\"next\"");
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+    Date date = new Date(System.currentTimeMillis() - DELTA);
+    String commitDate = DATE_FORMAT.format(date);
+    int pageNumber = FIRST_PAGE;
+
+    // TODO: The data gathering process here from GitHub API, will have to be merged with other
+    // related data providers. More information can be found in
+    // https://github.com/SAP/fosstars-rating-core/issues/69
+    try {
+      while (hasNextPage) {
+
```
artem-smotrakov: Please remove this extra line.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
```diff
@@ -0,0 +1,120 @@
...
+    when(provider.httpClient()).thenReturn(httpClient);
+
+    CloseableHttpResponse response = mock(CloseableHttpResponse.class);
+    when(httpClient.execute(any())).thenReturn(response);
+
+    HttpEntity entity = mock(HttpEntity.class);
+    when(response.getEntity()).thenReturn(entity);
+
+    try (InputStream content = getClass().getResourceAsStream("UsesSignedCommitTrue.json")) {
+
```
artem-smotrakov: Please remove this extra line.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
```diff
@@ -0,0 +1,120 @@
...
+      assertFalse(values.of(USES_VERIFIED_SIGNED_COMMITS).get().isUnknown());
+      assertEquals(USES_VERIFIED_SIGNED_COMMITS.value(true),
+          values.of(USES_VERIFIED_SIGNED_COMMITS).get());
+    }
+  }
+
+  // The test below works. Since the USES_VERIFIED_SIGNED_COMMITS value gets stored in cache due to
+  // first test. This will fail. Hence ignoring this.
+  @Test
+  @Ignore
```
artem-smotrakov: Please mock `AbstractGitHubDataProvider.cache()` method to return a dummy cache which doesn't store anything. Then, you can enable this test.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
```diff
@@ -0,0 +1,120 @@
...
+      assertEquals(USES_VERIFIED_SIGNED_COMMITS.value(false),
+          values.of(USES_VERIFIED_SIGNED_COMMITS).get());
+    }
+  }
+
+  // The test below works. Since the USES_VERIFIED_SIGNED_COMMITS value gets stored in cache due to
+  // first test. This will fail. Hence ignoring this.
+  @Test
+  @Ignore
+  public void testUsesSignedUknown() throws IOException {
```
artem-smotrakov: Typo: testUsesSignedUknown -> testUsesSignedUnknown

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
```diff
@@ -0,0 +1,120 @@
...
+    CloseableHttpClient httpClient = mock(CloseableHttpClient.class);
+    when(provider.httpClient()).thenReturn(httpClient);
+
+    CloseableHttpResponse response = mock(CloseableHttpResponse.class);
+    when(httpClient.execute(any())).thenReturn(response);
+
+    HttpEntity entity = mock(HttpEntity.class);
+    when(response.getEntity()).thenReturn(entity);
+
+    try (InputStream content = getClass().getResourceAsStream("UsesSignedCommitFalse.json")) {
```
artem-smotrakov: I don't see this file. Did you forget to add it?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+   * @return {@link BooleanValue} value of the feature {@link #USES_VERIFIED_SIGNED_COMMITS}.
+   */
+  private Value<Boolean> usesSignedCommits() {
+    double signedCounter = 0;
+    double counter = 0;
+    Date date = new Date(System.currentTimeMillis() - DELTA);
+    String commitDate = DATE_FORMAT.format(date);
+    int pageNumber = FIRST_PAGE;
+
+    // TODO: The data gathering process here from GitHub API, will have to be merged with other
```
artem-smotrakov: Minor: I suspect the comma needs to be removed.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,209 @@
...
+  CloseableHttpClient httpClient() {
+    return HttpClients.createDefault();
+  }
+
+  /**
+   * Check if next page is available, returns true. Otherwise the last page has been reached.
+   * 
+   * @param response {@link CloseableHttpResponse} from the GET Request.
+   * @return boolean if the header "link" contains 'rel="next"'
+   */
```
artem-smotrakov: Would you mind updating this to the following:

```
   * Check if the next page is available.
   * 
   * @param response {@link CloseableHttpResponse} from the GET Request.
   * @return True if the next page is available, false otherwise.
```

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommitTest.java
sourabhsparkala: Pushed it now.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,204 @@
...
+    String commitDate = DATE_FORMAT.format(date);
+    int pageNumber = FIRST_PAGE;
+    boolean hasNextPage = true;
+
+    // TODO: The data gathering process here from GitHub API will have to be merged with other
+    // related data providers. More information can be found in
+    // https://github.com/SAP/fosstars-rating-core/issues/69
+    try {
+      while (hasNextPage) {
+        CloseableHttpResponse response = commitInfo(commitDate, pageNumber);
```
artem-smotrakov: Please close the response with try-with-resources.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,204 @@
...
+        JsonNode commitNodeList = MAPPER.readTree(response.getEntity().getContent());
+        System.out.println(String.format("Page %s, Count %s", pageNumber, commitNodeList.size()));
+        if (commitNodeList == null || commitNodeList.size() == 0) {
+          break;
+        }
+
+        hasNextPage = hasNextPage(response);
```
artem-smotrakov: One minor thing. The variable doesn't seem to be really necessary. How about using a do-while loop to make it a bit shorter:

```java
do {
    CloseableHttpResponse response = commitInfo(commitDate, pageNumber);
    // do stuff
} while (hasNextPage(response));
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * Initializes a data provider.
+   *
+   * @param where A GitHub organization or user name.
+   * @param name A name of a repository.
+   * @param github An interface to the GitHub API.
+   * @param githubToken provided by user to access GitHub API.
+   */
+  public UsesSignedCommits(String where, String name, GitHub github, String githubToken)
+      throws IOException {
```
artem-smotrakov: The constructor doesn't seem to throw `IOException`.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,204 @@
...
+   * commits against the total number of commits in the year. If the density ratio in % crosses a
+   * certain percentage threshold ({@link #SIGNED_COMMIT_PERCENTAGE_THRESHOLD}), then it says that
+   * the project uses verified signed commits. Otherwise, it says that the project does not use
+   * verified signed commits.
+   * 
+   * @return {@link BooleanValue} value of the feature {@link #USES_VERIFIED_SIGNED_COMMITS}.
+   */
+  private Value<Boolean> usesSignedCommits() {
+    double signedCounter = 0;
+    double counter = 0;
```
artem-smotrakov: The counters should be integers.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -0,0 +1,233 @@
...
+    } catch (IOException e) {
+      System.out.printf("[x] Couldn't fetch data from api.github.com!%n");
+      System.out.printf("[x]     %s%n", e);
+      e.printStackTrace(System.out);
+      return UnknownValue.of(USES_VERIFIED_SIGNED_COMMITS);
+    }
+  }
+
+  /**
+   * Check if the project uses any signed verified commits.
```
artem-smotrakov: Please drop "any".

PR #70, opened 2020-03-19
Updated OssSecurityScore to use sub-scores

### LINKED ISSUE #24: Use LgtmScore in the open-source security rating
The `LgtmScore` is not currently used in the `OssSecurityRating`. Consider including the score to the `OssSecurityScore` directly, or to the `SecurityTestingScore`.

The `LgtmScore` may need to be adjusted, and test vectors for the score have to be added.

### LINKED ISSUE #67: Update OssSecurityScore to use sub-scores instead of features
Currently, the `OssSecurityScore` is a feature-based score. Since the `LgtmScore` is going to be added to the `OssSecurityScore` (see #24), the score needs to be updated to use sub-scores.

artem-smotrakov:

- Introduced a new `AverageCompositeScore` class which calculates an average of its sub-scores.
- Updated the `OssSecurityScore` class to extend `AverageCompositeScore` and include `DependencyScanScore` and `LgtmScore` sub-scores.
- Updated the demo tool to use `LgtmDataProvider`.
- Updated tests.

This closes #24 
This also closes #67


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
```diff
@@ -151,4 +180,33 @@ protected ScoreValue scoreValue(double value, Value... usedValues) {
...
+   *
+   * @param score The score.
+   * @param values The set of values.
+   * @return A value of the specified score.
+   * @throws IllegalArgumentException If a value for the score is not a {@link ScoreValue}.
+   */
+  static ScoreValue calculateIfNecessary(Score score, ValueHashSet values) {
+    Optional<Value> something = values.of(score);
+
+    // if the set of values doesn't contains a value for the specified score, then calculate it
```
sourabhsparkala: doesn't contains -> doesn't contain

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
```diff
@@ -151,4 +180,33 @@ protected ScoreValue scoreValue(double value, Value... usedValues) {
...
+    Value value = something.orElseGet(() -> UnknownValue.of(score));
+    if (value.isUnknown()) {
+      return score.calculate(values);
+    }
+
+    // if the set of values contains a value for the specified score, then return it
```
sourabhsparkala: doesn't contains -> doesn't contain

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
```diff
@@ -0,0 +1,120 @@
...
+   * Initializes a new {@link AverageCompositeScore} with a set of sub-scores.
+   *
+   * @param name A name of the score.
+   * @param subScores A set of sub-scores.
+   */
+  @JsonCreator
+  AverageCompositeScore(
+      @JsonProperty("name") String name,
+      @JsonProperty("subScores") Set<Score> subScores) {
+
```
sourabhsparkala: Please remove this line.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Calculate an overall score value as an average of the underlying sub-scores.
+   *
+   * @param values A number of values.
+   * @return An overall score.
+   */
+  @Override
+  public ScoreValue calculate(Value... values) {
+    ValueHashSet valueSet = new ValueHashSet(values);
+
```
sourabhsparkala: Can we remove this line as well?

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
```diff
@@ -0,0 +1,176 @@
...
+  private static final double PRECISION = 0.001;
+
+  @Test
+  public void smokeTest() {
+    TestAverageCompositeScore score = new TestAverageCompositeScore();
+    assertEquals(TestAverageCompositeScore.NAME, score.name());
+    assertEquals(2, score.allFeatures().size());
+    assertTrue(score.allFeatures().contains(FirstScore.FEATURE));
+    assertTrue(score.allFeatures().contains(SecondScore.FEATURE));
+    assertTrue(score.features().isEmpty());
```
sourabhsparkala: Please correct me if I am wrong. What I understand here is 
- score.allFeatures() indicate all features including sub-score features().
- score.features() indicates only the features of the root score, which is actually composed by sub-scores.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    double expectedScore = (firstValue + SecondScore.VALUE) / 2;
+    assertEquals(expectedScore, scoreValue.get(), PRECISION);
+
+    double expectedConfidence = (firstConfidence + Confidence.MAX) / 2;
+    assertEquals(expectedConfidence, scoreValue.confidence(), PRECISION);
+  }
+
+  private static class FirstScore extends FeatureBasedScore {
+
```
sourabhsparkala: Please remove this line.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      return obj.getClass() == getClass();
+    }
+  }
+
+  private static class SecondScore extends FeatureBasedScore {
+
```
sourabhsparkala: Please remove this line.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      return obj.getClass() == getClass();
+    }
+  }
+
+  private static class TestAverageCompositeScore extends AverageCompositeScore {
+
```
sourabhsparkala: Please remove this line.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScoreTest.java
```diff
@@ -25,10 +25,18 @@
 
 public class WeightedCompositeScoreTest {
 
+  private static final double PRECISION = 0.001;
```
sourabhsparkala: Other question: 
- Can this precision vary with time? Maybe after the inclusion of new features or the addition of new scores?
- Should we think about maybe having a range for this [0.01, 0.001].

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
artem-smotrakov: I would prefer to add an extra line when parameters of a method take multiple lines. The extra line helps to separate the method definition from the body.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
artem-smotrakov: Okay.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
artem-smotrakov: `Score.allFeatures()` returns all features which are used by a score. But It doesn't return sub-scores. `Score.features()` returns features which are directly used by a score.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
sourabhsparkala: ok

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
artem-smotrakov: I would prefer to keep this like since it separates the class definition and fields.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
sourabhsparkala: Got it

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScoreTest.java
artem-smotrakov: The precision depends on values which are used in the test. At the moment, the test uses `x.yz` doubles. The current precision should be enough to test those doubles. If the test starts using longer doubles, then the precision will need to be updated. I am not sure why we need a range for precisin.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
artem-smotrakov: I would prefer to keep this like since it separates the class definition and fields.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
artem-smotrakov: I would prefer to keep this like since it separates the class definition and fields.

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScoreTest.java
sourabhsparkala: I understood it wrong then, I thought the value is being compared with precision in difference. That is the expectedValue cannot be more or less than the actualValue by 0.001. Hence, asked if it makes sense to give a range or precision?

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
sourabhsparkala: ok

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
sourabhsparkala: ok

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScoreTest.java
sourabhsparkala: ok

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScoreTest.java
artem-smotrakov: > That is the expectedValue cannot be more or less than the actualValue by 0.001.

Yes, that's correct.

> Hence, asked if it makes sense to give a range or precision?

I am not sure how a range can help here. At the moment, doubles are compared like this:

```java
double actual = ...
double expected ...
assertEquals(expected, actual, PRECISION);
```

See also https://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertEquals(double,%20double,%20double)

Such a check ends up to something like this:

```java
if (abs(actual - expected) < PRECISION) {
    // throw an exception
}
```

I am not sure how a range of precision can make it better. Or, am I missing something?

artem-smotrakov: There is also `assertScore(Interval, ...)` method which checks that a score value belongs to a specific range. I think we need to use `assertEquals(double, double, double)` when we expect a specific score value, and `assertScore(Interval, ...)` method when we'd like to check if a score value belongs to a range.

artem-smotrakov: @sourabhsparkala

sourabhsparkala: I think this is an open discussion. As this is still not concrete. Shall we discuss this another time? For now, precision 0.001 is a really good benchmark.

artem-smotrakov: Sure.

artem-smotrakov:

Issue #72, opened 2020-03-24
Generate a page which contains ratings for well-known projects


artem-smotrakov:
Let's have a page which displays ratings for well-known projects. It would be useful for demo purposes, and also it can help to catch issues in rating calculation.

Here is a list of things to be done:

1. Write a tool to generate a report for multiple projects.
1. Define a list of projects.
1. Create a report for the defined projects.

PR #74, opened 2020-03-25
Implemented GitHubProjectFinder

artem-smotrakov:

- Added a new `GitHubProjectFinder` class for scanning organizations on GitHub.
- Added a new `SingleSecurityRatingCalculator` class which calculates a rating for a single project.
- Updated the `SecurityRatingCalculator` to use the new APIs.
- Added drafts of classes with will be used to finish #72 

This update contains multiple unfinished classes which will be finished later (implementation, javadoc, etc). That's just an initial design. Please focus on reviewing the `GitHubProjectFinder`  and `SingleSecurityRatingCalculator` classes and the design itself.


pom.xml
```diff
@@ -171,6 +172,11 @@
       <version>${version.mockito-core}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-yaml</artifactId>
```
sourabhsparkala: Looks like `jackson-dataformat-yaml` project would be deprecated in the [future](https://github.com/FasterXML/jackson-dataformat-yaml/blob/master/README.md).
Please update it to `jackson-dataformats-text`

artem-smotrakov: I think the repository is deprecated but the artifact `jackson-dataformats-yaml` is not. The code just moved to https://github.com/FasterXML/jackson-dataformats-text but the GAV stays the same.

https://github.com/FasterXML/jackson-dataformats-text/tree/master/yaml#maven-dependency

artem-smotrakov:

PR #75, opened 2020-03-25
Changed relative to absolute links to work with gh-pages

mibo:
Changed relative to absolute links to work with gh-pages for the source code samples.

Is it required to create an issue for this minor contribution?


artem-smotrakov:
> Is it required to create an issue for this minor contribution?

Let's not require that but I opened #76

artem-smotrakov:
This fixes #76

Issue #35, opened 2020-02-21
Update some classes to use Logger


Sachpat:
There are some classes which contain Sysouts. They should better use a Logger if possible.

artem-smotrakov:
```
$ grep -r "System.out" src/
src/main/java/com/sap/sgs/phosphor/fosstars/data/Terminal.java:    System.out.print(">>> ");
src/main/java/com/sap/sgs/phosphor/fosstars/data/Terminal.java:    System.out.println(question);
src/main/java/com/sap/sgs/phosphor/fosstars/data/Terminal.java:    System.out.print(">>> ");
src/main/java/com/sap/sgs/phosphor/fosstars/data/Terminal.java:    System.out.println(phrase);
src/main/java/com/sap/sgs/phosphor/fosstars/data/IsEclipse.java:    System.out.println("[+] Figuring out if the project belongs to the Eclipse Software Foundation ...");
src/main/java/com/sap/sgs/phosphor/fosstars/data/lgtm/LgtmDataProvider.java:      System.out.printf("[x] Couldn't fetch data from lgtm.com!%n");
src/main/java/com/sap/sgs/phosphor/fosstars/data/lgtm/LgtmDataProvider.java:      System.out.printf("[x]     %s%n", e);
src/main/java/com/sap/sgs/phosphor/fosstars/data/lgtm/LgtmDataProvider.java:    System.out.printf("[+] check project: https://github.com/%s/%s%n", where, name);
src/main/java/com/sap/sgs/phosphor/fosstars/data/lgtm/LgtmDataProvider.java:    System.out.printf("[+] Uses LGTM:   %s%n", values.of(USES_LGTM).orElse(UnknownValue.of(USES_LGTM)));
src/main/java/com/sap/sgs/phosphor/fosstars/data/lgtm/LgtmDataProvider.java:    System.out.printf("[+] Worse grade: %s%n", values.of(WORSE_LGTM_GRADE).orElse(UnknownValue.of(WORSE_LGTM_GRADE)));
src/main/java/com/sap/sgs/phosphor/fosstars/data/ValueCache.java:      System.out.println("[+] The default value cache doesn't exist yet.");
src/main/java/com/sap/sgs/phosphor/fosstars/data/ValueCache.java:      System.out.println("[!] Could not load the default value cache!");
src/main/java/com/sap/sgs/phosphor/fosstars/data/ValueCache.java:      e.printStackTrace(System.out);
[...]

PR #80, opened 2020-03-26
Updated the demo tool to handle multiple projects

### LINKED ISSUE #10: SecurityRatingCalculatorTest should not depend on GitHub API
The test fails when the GitHub API rate limit exceeded or some networking issues occur. That's why the test is currently disabled.

The test needs to be re-written to avoid connecting to the GitHub API.

### LINKED ISSUE #79: RatingValue should be serializable to JSON
The `RatingValue` class needs to be updated to support serialization with Jackson.

artem-smotrakov:
This is the next update for #72 

- Implemented the `MultipleSecurityRatingsCalculator` class.
- Introduced a new option `--config` which allows setting a configuration.
- Updated `SecurityRatingCalculator` to use `GitHubProjectFinder` and `MultipleSecurityRatingsCalculator` if the `--config` option is set.
- Re-worked error handling in the `SecurityRatingCalculator` class.
- Make the `RatingValue` class serializable via Jackson. This fixes #79 
- Updated the tests.
- Fixed #10

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
@@ -1,0 +1,68 @@
+package com.sap.sgs.phosphor.fosstars.tool.github;
+
+import com.sap.sgs.phosphor.fosstars.data.NoUserCallback;
+import com.sap.sgs.phosphor.fosstars.data.UserCallback;
+import java.io.IOException;
+import java.util.Objects;
+import org.kohsuke.github.GitHub;
+
+/**
+ * A base class for rating calculators.
+ */
+abstract class AbstractRatingCalculator {
+
+  /**
+   * An interface to GitHub.
+   */
+  final GitHub github;
+
+  /**
+   * A token for accessing the GitHub APIs.
+   */
+  String token;
+
+  /**
+   * An interface for interacting with a user.
+   */
+  UserCallback callback = NoUserCallback.INSTANCE;
+
+  /**
+   * Initializes a new calculator.
+   *
+   * @param github An interface to GitHub.
+   */
+  AbstractRatingCalculator(GitHub github) {
+    this.github = Objects.requireNonNull(github, "Oh no! An interface to GitHub can't be null!");
+  }
+
+  /**
+   * Sets a token for accessing the GitHub APIs.
+   *
+   * @param token The token.
+   * @return The same calculator.
+   */
+  AbstractRatingCalculator token(String token) {
+    this.token = token;
+    return this;
+  }
+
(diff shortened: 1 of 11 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * An interface to GitHub.
+   */
+  final GitHub github;
+
+  /**
+   * A token for accessing the GitHub APIs.
+   */
+  String token;
```
Sachpat: We don't need any scope for this token variable?

artem-smotrakov: The field is used by classes which extend `AbstractRatingCalculator`. Technically, the field could have `protected` score. However, the classes which extend `AbstractRatingCalculator` are in the same package, so the default scope can be used as well. Since the default scope is more restrictive than `protected`, I would prefer to use the default scope.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
```diff
@@ -27,33 +40,59 @@
...
    * Entry point.
+   *
+   * @param args Command-line parameters.
+   */
+  public static void main(String... args) {
+    try {
+      run(args);
+    } catch (Exception e) {
+      System.out.printf("[x] Something went wrong!%n");
+      e.printStackTrace(System.out);
```
Sachpat: e.printstacktrace() by default prints it to System.out isnt it?

artem-smotrakov: Nope, it uses `System.err` by default. In fact, I am planning to update it to use loggers, see #35.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
Sachpat: Ok, thanks for the explanation. 
> However, the classes which extend AbstractRatingCalculator are in the same package,
Hmm no plans of allowing the abstract class to be extended in other packages in future?

artem-smotrakov: No plans at the moment. Right now it's not public API, the `AbstractRatingCalculator` class is just used to store some common code.

artem-smotrakov:
> Looks good to me. Thanks for the explanation for the questions.

PR #81, opened 2020-03-30
Updated the demo tool to generate a report for multiple projects

artem-smotrakov:

- Implemented `MergedJsonReporter` which stores info about processed projects to a JSON file.
- Implemented `MarkdownReporter` which generates a markdown report for processed projects.
- Implemented a cache of processed projects.
- Updated the demo tool to generate a markdown report for multiple projects.
- Created a draft report in `docs/oss/security` directory. The report contains a few projects, it'll be extended later.
- Updated tests.

This is a part of #72

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
@@ -1,7 +1,80 @@
 package com.sap.sgs.phosphor.fosstars.tool.github;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sap.sgs.phosphor.fosstars.tool.Project;
 import com.sap.sgs.phosphor.fosstars.tool.Reporter;
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
-abstract class AbstractReporter implements Reporter {
+abstract class AbstractReporter<T extends Project> implements Reporter<T> {
+
+  /**
+   * For serialization.
+   */
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  /**
+   * A type reference of a list of {@link GitHubProject}s for deserialization.
+   */
+  private static final TypeReference<List<GitHubProject>> LIST_OF_GITHUB_PROJECTS_TYPE
+      = new TypeReference<List<GitHubProject>>() {};
+
+  /**
+   * Merges two lists of projects. The merge is based on URLs.
+   * If there are two projects with the same URLs,
+   * then the method uses the project from the first list.
+   *
+   * @param projects The first list of projects.
+   * @param extraProjects The second list of projects.
+   * @return A list of projects which contains all projects from the original lists.
+   */
+  static List<GitHubProject> merge(
+      List<GitHubProject> projects, List<GitHubProject> extraProjects) {
+
+    Map<URL, GitHubProject> map = new HashMap<>();
+    for (GitHubProject extraProject : extraProjects) {
+      map.put(extraProject.url(), extraProject);
+    }
+    for (GitHubProject project : projects) {
+      map.put(project.url(), project);
+    }
+
+    List<GitHubProject> allProjects = new ArrayList<>();
+    for (Map.Entry<URL, GitHubProject> entry : map.entrySet()) {
+      allProjects.add(entry.getValue());
+    }
+
+    return allProjects;
+  }
+
+  /**
+   * Loads projects from a JSON file.
+   * If the file doesn't exist, then the method returns an empty list.
+   *
+   * @param extraSourceFileName A path to the file.
+   * @return A list of loaded extra projects.
+   * @throws IOException If the projects couldn't be loaded.
+   */
+  static List<GitHubProject> loadProjects(String extraSourceFileName) throws IOException {
+    if (extraSourceFileName == null) {
+      return Collections.emptyList();
+    }
+    Path path = Paths.get(extraSourceFileName);
+    if (!Files.exists(path)) {
+      System.out.printf(
+          "[!] Oh no! I could not load extra projects from %s%n", extraSourceFileName);
+      return Collections.emptyList();
+    }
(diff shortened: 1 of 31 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
```diff
@@ -1,7 +1,80 @@
...
+  /**
+   * Merges two lists of projects. The merge is based on URLs.
+   * If there are two projects with the same URLs,
+   * then the method uses the project from the first list.
+   *
+   * @param projects The first list of projects.
+   * @param extraProjects The second list of projects.
+   * @return A list of projects which contains all projects from the original lists.
+   */
+  static List<GitHubProject> merge(
```
Sachpat: Doesn't GitHubProject have a equals method? with that, we can simply handle this with 
`
for (Object x : two){
   if (!one.contains(x))
      one.add(x);
}
`
?

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
```diff
@@ -192,6 +208,89 @@ private static void processConfig(String filename, GitHub github, String githubT
...
+    Objects.requireNonNull(config, "Oh no! Config is null!");
+    if (config.reportConfigs == null) {
+      return Collections.singletonList(Reporter.DUMMY);
+    }
+
+    List<Reporter<GitHubProject>> reporters = new ArrayList<>();
+    for (ReportConfig reportConfig : config.reportConfigs) {
+      Objects.requireNonNull(reportConfig.type, "Hey! Reporter type can't be null!");
+      switch (reportConfig.type) {
+        case "markdown":
```
Sachpat: possible to extract these "markdown" , "json" to constants.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
artem-smotrakov: Yes, `GitHubProject` has `equals()` and `hashCode()` methods. Besides project and organization names, the class  also contains a rating value. The rating value should not be taken into account in the `merge()` method. However, the current implementations of `equals()` and `hashCode()` do not use the rating value. So, your suggestion would work. If `equals()` and `hashCode()` method change in the future, the `merge()` method may stop working as expected. I would like to avoid that, so that I implemented it with an extra map. Please let me know if you see anything wrong here.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
artem-smotrakov: Yup, I thought about introducing a enum. I think it would be even better.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
Sachpat: Makes sense @artem-smotrakov thanks for the information.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
Sachpat: Yes.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
```diff
@@ -1,7 +1,80 @@
...
+
+  /**
+   * Loads projects from a JSON file.
+   * If the file doesn't exist, the the method returns an empty list.
+   *
+   * @param extraSourceFileName A path to the file.
+   * @return A list of loaded extra projects.
+   * @throws IOException If the projects couldn't be loaded.
+   */
+  static List<GitHubProject> loadProjects(String extraSourceFileName) throws IOException {
```
sourabhsparkala: Please consider this
- If the files are not found? Then we return an empty list.
- Why not catch the IOException also and then return the emptyList?

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/MarkdownReporter.java
```diff
@@ -1,5 +1,184 @@
...
+  /**
+   * A template for a table row in the report.
+   */
+  private static final String PROJECT_LINE_TEMPLATE
+      = "| %URL% | %SCORE% | %LABEL% | %CONFIDENCE% | %DATE% |";
+
+  /**
+   * A date formatter.
+   */
+  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("MMM d, YYYY");
```
sourabhsparkala: Please consider using default Locale.US here

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/MarkdownReporter.java
```diff
@@ -1,5 +1,184 @@
...
+
+  /**
+   * Prepares a description how a rating was calculated for a project.
+   *
+   * @param project The project.
+   * @return The details of the rating calculation.
+   */
+  private static String detailsOf(GitHubProject project) {
+    if (!project.ratingValue().isPresent()) {
+      return "unknown";
```
sourabhsparkala: looks like a static final variable

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractReporter.java
artem-smotrakov: The method tries to load a file only if a file name is specified (not null) and the file exists. However, if the file exists, the method makes sure that the content of the file is correct. That's why it doesn't return an empty list if deserialization throws an IOException.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/MarkdownReporter.java
artem-smotrakov: Yup, that's right.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/MarkdownReporter.java
artem-smotrakov: I can do that. Most likely the report is going to be made nicer, so this `static final` will likely be removed.

artem-smotrakov:

PR #83, opened 2020-03-30
Updated Jackson to 2.10.3

### LINKED ISSUE #73: Update Jackson components to the latest versions
The project uses Jackson libraries of version `2.10.1`. However, never versions are available. We need to update to the latest version.

artem-smotrakov:
This fixes #73

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,8 +10,8 @@
   <name>Fosstars Rating Core</name>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <version.jackson-databind>2.9.0</version.jackson-databind>
-    <version.jackson-dataformat-yaml>2.9.0</version.jackson-dataformat-yaml>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -10,8 +10,8 @@
   <name>Fosstars Rating Core</name>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <version.jackson-databind>2.9.0</version.jackson-databind>
-    <version.jackson-dataformat-yaml>2.9.0</version.jackson-dataformat-yaml>
+    <version.jackson-databind>2.10.3</version.jackson-databind>
+    <version.jackson-dataformat-yaml>2.10.3</version.jackson-dataformat-yaml>
```
Sachpat: Hmm, do we actually need two properties ? we can manage with a single one right ? ;)

artem-smotrakov: We can. Let's keep two properties since the list of properties shows all dependencies in a shorter way.

PR #89, opened 2020-04-01
Calculated ratings for Apache and Eclipse projects

### LINKED ISSUE #72: Generate a page which contains ratings for well-known projects

artem-smotrakov:
You can find the report at https://github.com/artem-smotrakov/fosstars-rating-core/tree/generate-report/docs/oss/security

This closes #72


src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/MarkdownReporter.java
```diff
@@ -98,7 +99,7 @@ public void runFor(List<GitHubProject> projects) throws IOException {
     for (GitHubProject project : allProjects) {
       stars.put(project, starsOf(project));
     }
-    allProjects.sort(Comparator.comparingInt(stars::get));
+    allProjects.sort(Collections.reverseOrder(Comparator.comparingInt(stars::get)));
```
sourabhsparkala, line 102: allProjects.sort(Comparator.comparingInt(stars::get, Comparator.reverseOrder())); would this be better? 

- I feel like Collections.reverseOrder() would have O(n^2) and Comparator.comparingInt() another O(n^2). Maybe I am wrong.

artem-smotrakov: > * I feel like Collections.reverseOrder() would have O(n^2) and Comparator.comparingInt() another O(n^2). Maybe I am wrong.

I am not sure about that. I think `Collections.reverseOrder()` just inverse the comparator. In other words, just changes the sign of value returned by the wrapped comparator. That should not increase complexity.

sourabhsparkala: ok. good to know.

artem-smotrakov:
> * What is the purpose of this report being published within the project?

Please see the description of #72. Without compiling and running the tool, one may have a quick look at the report to see how the rating calculating works.

> * Should we keep the code and the reports separate?

I think it's fine to keep it in one place. At the moment, I don't see any better place for this report.

> * What is the best way to handle this now? Should we ignore this folder when pulling or cloning the code?

You can't ignore the folder when you pull the code. Fetching the repository with the report takes a few seconds.

sourabhsparkala:
Ok. Thank you for the clarification.

PR #105, opened 2020-04-02
Store test vectors in YAML

### LINKED ISSUE #104: Consider using YAML for storing test vectors
Currently, test vectors are defined in the Jupyter notebooks and stored to CSV files. When someone updates test vectors in the notebooks, it usually results to many unrelated updates. That makes it difficult to review the changes in the test vectors. The CSV files are also difficult to review. 

We need to come up with a new way how test vectors may be easily defined, updated and reviewed. One of the possible solutions is to store them in YAML files because YAML is more human-readable rather than CSV or JSON.

However, we need to keep the notebooks since the provide a good way to build visualisations and statistics for the test vectors.

artem-smotrakov:

- Stored the test vectors in YAML files.
- Removed the CSV files with test vectors.
- Updated the verification procedures to load test vectors from YAML files.
- Removed definitions of test vectors from the Jupyter notebooks.
- Set date formats for serialization.
- Updated tests.

This fixes #104


artem-smotrakov:
> @artem-smotrakov Just a quick question, why did we decide to move from csv to yaml file? for the ease of use.

To make them easier to review when they get updated. At the moment, the test vectors are defined in the Jupyter notebooks and then stored in CSV file. Both are difficult to review in a pull request.

artem-smotrakov:
> @artem-smotrakov Just a quick question, why did we decide to move from csv to yaml file? for the ease of use.

Please also see the description of #104 which is addressed in this pull request.

src/test/java/com/sap/sgs/phosphor/fosstars/model/value/DateValueTest.java
```diff
@@ -53,6 +53,7 @@ public void serialization() throws IOException {
     ObjectMapper mapper = new ObjectMapper();
     DateValue dateValue = new DateValue(new DateFeature("test"), new Date());
     byte[] bytes = mapper.writeValueAsBytes(dateValue);
+    System.out.println(new String(bytes));
```
Sachpat, line 56: do you still need it? I guess you added this for yourself right? :)

Sachpat:
> > @artem-smotrakov Just a quick question, why did we decide to move from csv to yaml file? for the ease of use.
> 
> Please also see the description of #104 which is addressed in this pull request.

Thanks I had a look, yes it makes sense to have yaml file.

src/test/java/com/sap/sgs/phosphor/fosstars/model/value/DateValueTest.java
artem-smotrakov: Good catch! I was going to remove it but I forgot.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -50,11 +51,13 @@
   /**
    * A date when the vulnerability was introduced.
    */
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd", locale = "us_US")
```
sourabhsparkala, line 54: why not pattern yyyy-MM-dd'T'HH:mm:ss.SSSZ here?

artem-smotrakov: I think that `yyyy-MM-dd` is enough for a vulnerability because it is not necessary to know exact time when it was introduced or fixed. Then, `yyyy-MM-dd` is easier to read.

src/main/jupyter/oss/security/commons.py
```diff
@@ -129,3 +15,42 @@ def draw_hists(columns, test_vectors, width=20, height=5):
...
+                elif feature_type == 'DateValue':
+                    feature_value = value['date']
+                elif feature_type == 'IntegerValue' or feature_type == 'PositiveIntegerValue':
+                    feature_value = value['number']
+                elif feature_type == 'LgtmGradeValue':
+                    feature_value = value['value']
+                elif feature_type == 'UnknownValue':
+                    feature_value = 'unknown'
+                elif feature_type == 'VulnerabilitiesValue':
+                    feature_value = '...'
```
sourabhsparkala, line 43: is this a place holder?

src/test/java/com/sap/sgs/phosphor/fosstars/model/value/DateValueTest.java
```diff
@@ -62,4 +62,4 @@ public void serialization() throws IOException {
     assertEquals(dateValue.hashCode(), clone.hashCode());
   }
 
-}
\ No newline at end of file
+}
```
sourabhsparkala, line 65: Is there any change in this file?

src/main/jupyter/oss/security/commons.py
artem-smotrakov: Yes. For complex type such as `VulnerabilitiesValue`, the function returns "...". Currently I don't know a good way how to visualise info about vulnerabilities in the notebooks.

src/test/java/com/sap/sgs/phosphor/fosstars/model/value/DateValueTest.java
artem-smotrakov: There used to be changes but not anymore. I'll revert this.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
sourabhsparkala: Agree

artem-smotrakov:
> Minor comments.
> 
> Everything else looks good

src/main/jupyter/oss/security/commons.py
sourabhsparkala: ok

PR #109, opened 2020-04-06
Updated the demo tool and data providers to use loggers

### LINKED ISSUE #35: Update some classes to use Logger

artem-smotrakov:

- Updated Log4j to 2.13.1
- Replaced System.out with loggers
- Updated the config for Log4j

This fixes #35

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
@@ -84,7 +84,7 @@ public class UsesSignedCommits extends AbstractGitHubDataProvider {
   @Override
   public UsesSignedCommits update(ValueSet values) throws IOException {
     Objects.requireNonNull(values, "Hey! Values can't be null!");
-    System.out.println("[+] Figuring out if the project uses verified signed commits ...");
+    logger.info("Figuring out if the project uses verified signed commits ...");
 
     Optional<Value> something = cache().get(url, USES_VERIFIED_SIGNED_COMMITS);
     if (something.isPresent()) {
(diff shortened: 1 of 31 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -113,8 +113,7 @@ public UsesSignedCommits update(ValueSet values) throws IOException {
     try {
       return new BooleanValue(USES_VERIFIED_SIGNED_COMMITS, askGithub());
     } catch (IOException e) {
-      System.out.printf("[x] Couldn't fetch data from api.github.com!%n");
-      System.out.printf("[x]     %s%n", e);
+      logger.error("Couldn't fetch data from api.github.com!", e);
```
Sachpat, line 116: Hmm why logger.error only here and logger.warn at other places for the exception? :)

Sachpat: For e.g., here https://github.com/SAP/fosstars-rating-core/pull/109/files#diff-93fae7ef1f1df793b699455448ed0af8R89

artem-smotrakov: There is no reason for that. That just happened and we didn't pay attention since it makes no harm. I'll update the code to use `warn()` to make it consistent (except the `main()` method).

Sachpat: Cool thanks :)

artem-smotrakov:
> Looks good to me.

PR #114, opened 2020-04-08
Introduced a feature for Dependabot

### LINKED ISSUE #94: Introduce a new feature for Dependabot
[Dependabot](https://dependabot.com/) offers automatic dependency updates. In particular, when a vulnerability in dependencies is detected, it opens a pull request to update the vulnerable dependency to the safe version.

A new boolean feature may be added to show if a project takes advantage of Dependabot and merges pull requests which it creates. The new feature may then contribute to the dependency scan score or somewhere else.

Please also implement a data provider (or update the existing ones) to gather the feature for projects on GitHub.

artem-smotrakov:
Here is what was done:

- Added a new `OssFeatures.USES_DEPENDABOT` feature.
- Added a new `UsesDependabot` data provider.
- Added `UsesDependabotTest`

This closes #94

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesDependabot.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesDependabot.java
@@ -1,0 +1,144 @@
+package com.sap.sgs.phosphor.fosstars.data.github;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_DEPENDABOT;
+
+import com.sap.sgs.phosphor.fosstars.model.Value;
+import com.sap.sgs.phosphor.fosstars.model.ValueSet;
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Date;
+import java.util.Objects;
+import java.util.Optional;
+import org.kohsuke.github.GHCommit;
+import org.kohsuke.github.GHContent;
+import org.kohsuke.github.GHRepository;
+import org.kohsuke.github.GitHub;
+
+/**
+ * <p>This data provider checks if an open-source project on GitHub
+ * uses Dependabot, and fills out the
+ * {@link com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures#USES_DEPENDABOT} feature.</p>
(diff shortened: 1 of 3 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesDependabot.java
```diff
@@ -0,0 +1,148 @@
...
+   * @return True if a commit from Dependabot was found, false otherwise.
+   */
+  private boolean hasDependabotCommits() {
+    try {
+      GHRepository repository = github.getRepository(path);
+
+      Date date = Date.from(Instant.now().minus(LATEST_MONTHS));
+      for (GHCommit commit : repository.listCommits().asList()) {
+        if (commit.getCommitDate().before(date)) {
+          break;
```
Sachpat, line 81: Hmm shouldn't it be continue instead of break here?  If I understand correctly, you dont want to check for a commit if its older than few months right?

artem-smotrakov: The method takes into account only commits that were made last year. It assumes that the commits is sorted by commit dates (which is a natural order for in commit history). If it finds a commit which is older than the date, then it breaks because other commits are going to be even older.

Let me check if it's a correct assumption.

artem-smotrakov: The docs say nothing about how `listCommits()` sorts the commits. However, I found out that the GitHub API offers `since` parameter. I think it's better to use it rather than make an assumption about commits order. I'll update the code to use `GHCommitQueryBuilder`.

https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHCommitQueryBuilder.html

Sachpat: Yes, this makes sense to me.

artem-smotrakov: @Sachpat I've updated the code. Please take a look.

artem-smotrakov:
> Looks good to me now.

PR #112, opened 2020-04-07
Extended DataProvider interface

artem-smotrakov:
Although the changeset contains quite a lot of updates, most of the them just refactor the existing functionality. The current behavior doesn't change much.

Here is a list of main updates:

- Added new methods to the `DataProvider` interface.
  Data providers now gathers data about specific objects
  such as open-source projects.
- Added a new `AbstractProvider` class.
- Added an interface for caches of feature values.
- Added new `StandardValueCache` and `GitHubProjectValueCache` classes.
- Updated the demo tool to manage common value cache.
- Enhanced and simplified the existing data providers.
- Updated tests.

This closes #111

PR #119, opened 2020-04-09
Introduced a feature for package managers

### LINKED ISSUE #116: Introduce a feature for package manager
GitHub can send a security alert to a project if a vulnerability is detected in one of its dependencies. This info could be used in the dependency scan score. However, GitHub supports only specific package managers:

https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies

https://help.github.com/en/github/visualizing-repository-data-with-graphs/listing-the-packages-that-a-repository-depends-on

Therefore, the score needs to know if a project uses one of the supported package managers. Let's introduce a feature which holds info about package manager.

artem-smotrakov:

- Added new `PackageManager`, `PackageManagers` and `PackageManagersValue` classes.
- Added a new `OssFeatures.PACKAGE_MANAGERS` feature.
- Added new string matchers.
- Added a new `PackageManagement` data provider.
- Added new tests.

This closes #116

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
@@ -1,0 +1,232 @@
+package com.sap.sgs.phosphor.fosstars.data.github;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.LANGUAGES;
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.PACKAGE_MANAGERS;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.C_SHARP;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.F_SHARP;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.JAVA;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.JAVASCRIPT;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.PHP;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.PYTHON;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.RUBY;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.SCALA;
+import static com.sap.sgs.phosphor.fosstars.model.value.Language.VISUALBASIC;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.COMPOSER;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.DOTNET;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.GRADLE;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.MAVEN;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.NPM;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.PIP;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.RUBYGEMS;
+import static com.sap.sgs.phosphor.fosstars.model.value.PackageManager.YARN;
+
+import com.sap.sgs.phosphor.fosstars.model.Value;
+import com.sap.sgs.phosphor.fosstars.model.ValueSet;
+import com.sap.sgs.phosphor.fosstars.model.value.Language;
+import com.sap.sgs.phosphor.fosstars.model.value.Languages;
+import com.sap.sgs.phosphor.fosstars.model.value.PackageManager;
+import com.sap.sgs.phosphor.fosstars.model.value.PackageManagers;
+import com.sap.sgs.phosphor.fosstars.model.value.ValueHashSet;
+import java.io.IOException;
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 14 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
```diff
@@ -0,0 +1,232 @@
...
+   * Maps a package manager to a list of its possible config files.
+   *
+   * @see <a href="https://help.github.com/en/github/visualizing-repository-data-with-graphs/listing-the-packages-that-a-repository-depends-on">
+   *      Listing the packages that a repository depends on</a>
+   */
+  private static final Map<PackageManager, Predicate<String>[]> CONFIG_FILES_PATTERNS
+      = new EnumMap<>(PackageManager.class);
+
+  static {
+    register(MAVEN, "pom.xml"::equals);
```
Sachpat, line 78: Are these not buildDescriptors? my understanding is package managers are something more than just pom.xmls or something. please correct me if I am wrong.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
```diff
@@ -0,0 +1,232 @@
...
+  static {
+    register(JAVA, MAVEN, GRADLE);
+    register(SCALA, MAVEN);
+    register(JAVASCRIPT, NPM, YARN);
+    register(C_SHARP, DOTNET);
+    register(F_SHARP, DOTNET);
+    register(VISUALBASIC, DOTNET);
+    register(PYTHON, PIP);
+    register(RUBY, RUBYGEMS);
+    register(PHP, COMPOSER);
```
Sachpat, line 65: If you were looking for some package managers for Go https://github.com/golang/go/wiki/PackageManagementTools

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
```diff
@@ -0,0 +1,232 @@
...
+
+  /**
+   * Adds a number of known package managers for a language.
+   *
+   * @param language The language.
+   * @param list The package managers.
+   */
+  private static void register(Language language, PackageManager... list) {
+    if (KNOWN_PACKAGE_MANAGERS.containsKey(language)) {
+      throw new IllegalArgumentException(String.format("%s is already there!", language));
```
Sachpat, line 96: Hmm instead of throwing the exception straight away, how about checking if it's a new packagemanager not associated with the language yet?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
artem-smotrakov: Good to know. In fact, the data provider don't cover all existing package managers. Currently it focuses only on those which are supported by the security alerts on GitHub. That's going to be used in the dependency scans score in the future. Unfortunately, the security alerts don't support Go.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
artem-smotrakov: This check is done only once when the class is loaded. I would prefer to keep it as simple as possible and allow specifying a list of package managers for a language only once.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/PackageManagement.java
```diff
@@ -0,0 +1,232 @@
...
+
+  /**
+   * Gets a list of package managers that are used in a project.
+   *
+   * @return A value of the feature
+   *         {@link com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures#PACKAGE_MANAGERS}.
+   * @throws IOException If something went wrong.
+   */
+  private Value<PackageManagers> packageManagers() throws IOException {
+    Optional<Value> something = cache().get(url, PACKAGE_MANAGERS);
```
Sachpat, line 141: Where is url defined? I couldn't find it. or am I missing something?

Sachpat: oh it's in AbstractGitHubDataProvider?

artem-smotrakov: The `url` field is in the `AbstractGitHubDataProvider` class. This field is going to be removed in #112

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/PackageManagers.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A set of package managers.
+ */
+public class PackageManagers {
```
Sachpat, line 17: Why do we need this class?  Set<PackageManager> or List<PackageManager> would be the same right?

artem-smotrakov: Right, that's just a set. I am even thinking about implementing a set interface. I would prefer to keep this class (and other similar ones such as `Languages`). The class allows adding convenient methods if necessary, and also using `Value<PackageManagers` instead of `Value<Set<PackageManager>>` which is not a big deal but looks nicer to me.

artem-smotrakov:
> LGTM.

PR #120, opened 2020-04-09
Separate interactive data providers

### LINKED ISSUE #108: Add DataProvider.interactive() method
It would be good if a data provider can tell whether it interacts with a user or not. Let's add a new `DataProvider.interactive()` method which returns true if a provider talks to a user, and false otherwise. Some of the existing data providers, which can talk to a user, will need to be split to two.

artem-smotrakov:

- Added a new `DataProvider.interactive()` method.
- Extracted interactive part of the existing data providers to new `AskAboutUnpatchedVulnerabilities` and `AskAboutSecurityTeam` data providers.
- Added tests.
- Updated the demo tool.

This closes #108

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
@@ -1,0 +1,39 @@
+package com.sap.sgs.phosphor.fosstars.data.interactive;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.HAS_SECURITY_TEAM;
+
+import com.sap.sgs.phosphor.fosstars.model.ValueSet;
+import com.sap.sgs.phosphor.fosstars.model.value.BooleanValue;
+import com.sap.sgs.phosphor.fosstars.tool.YesNoSkipQuestion;
+import com.sap.sgs.phosphor.fosstars.tool.YesNoSkipQuestion.Answer;
+
+/**
+ * This data provider asks a user about security teams.
+ */
+public class AskAboutSecurityTeam<T> extends AbstractInteractiveDataProvider<T> {
+
+  @Override
+  protected AskAboutSecurityTeam<T> ask(T object, ValueSet values) {
+    String question = String.format(
+        "Does project '%s' have a security team? Say yes, no, or skip, please.", object);
+
+    Answer answer = new YesNoSkipQuestion(callback, question).ask();
+    switch (answer) {
+      case YES:
+        values.update(new BooleanValue(HAS_SECURITY_TEAM, true));
+        break;
+      case NO:
+        values.update(new BooleanValue(HAS_SECURITY_TEAM, false));
+        break;
+      case SKIP:
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Hey! You gave me an unexpected answer: %s", answer));
+    }
+
+    // TODO: store the answer in the SecurityTeamStorage
+
+    return this;
+  }
+}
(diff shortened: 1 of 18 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
```diff
@@ -0,0 +1,37 @@
...
+        break;
+      case NO:
+        values.update(new BooleanValue(HAS_SECURITY_TEAM, false));
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Hey! You gave me an unexpected answer: %s", answer));
+    }
+
+    // TODO: store the answer in the SecurityTeamStorage
```
sourabhsparkala, line 33: Is there an issue tracking this?

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * This data provider asks a user about security teams.
+ */
+public class AskAboutSecurityTeam<T> extends AbstractInteractiveDataProvider<T> {
+
+  @Override
+  protected AskAboutSecurityTeam<T> ask(T object, ValueSet values) {
+    String question = String.format(
+        "Does project '%s' have a security team? Say yes, no, or skip, please.", object);
```
sourabhsparkala, line 18: How is skip handled here? I mean in the switch case?

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutUnpatchedVulnerabilities.java
```diff
@@ -0,0 +1,65 @@
...
+        unpatchedVulnerabilities.add(vulnerability);
+        answer = new YesNoQuestion(callback, "One more?").ask();
+      } while (answer == YES);
+    }
+
+    Value<Vulnerabilities> vulnerabilities = knownVulnerabilities(values);
+    vulnerabilities.get().add(unpatchedVulnerabilities);
+    values.update(vulnerabilities);
+
+    // TODO: store the entered vulnerabilities somewhere
```
sourabhsparkala, line 49: Is there an issue tracking this?

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
```diff
@@ -0,0 +1,37 @@
...
+        values.update(new BooleanValue(HAS_SECURITY_TEAM, false));
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Hey! You gave me an unexpected answer: %s", answer));
+    }
+
+    // TODO: store the answer in the SecurityTeamStorage
+
+    return null;
```
sourabhsparkala, line 35: `return this;` ?

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
artem-smotrakov: No, we don't have an issue for that. To be honest, I am not sure if it should be implemented. The interactive providers are not a priority for now.

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
artem-smotrakov: Good point. I'll add a case for `SKIP`.

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutUnpatchedVulnerabilities.java
artem-smotrakov: No, we don't have an issue for that. To be honest, I am not sure if it should be implemented. The interactive providers are not a priority for now.

src/main/java/com/sap/sgs/phosphor/fosstars/data/interactive/AskAboutSecurityTeam.java
artem-smotrakov: Yup!

artem-smotrakov:
Thanks for the review @sourabhsparkala ! I've addressed your comments.

PR #126, opened 2020-04-16
Introduced not-applicable values

### LINKED ISSUE #97: Introduce a special value Not Applicable
It may happen that some features or scores doesn't make much sense for a particular open-source project. For example, if a project doesn't contains C/C++ code, then it's unlikely that it has memory corruption issues. Therefore, for such a project, it doesn't makes much sense to use a score which assesses memory safety. On the other hand, such a score may be extremely important for projects which contain C/C++ code.

At the moment, if a score aggregates values from a number of sub-scores, it is not possible to skip the sub-scores which don't make much sense for a specific project. In order to make it possible, let's introduce a new special score value "Not Applicable". The special value means that a score value should be ignored by a high-level score.

Currently it's not clear yet if a feature value should also support it.

The scores which uses other sub-scores need to be updated to expect the new special value.

artem-smotrakov:

- Added a new `Value.isNotApplicable()` method.
- Added a new `NotApplicableValue` class.
- Updated implementations of the `Value` interface to implement the new method.
- Added a new `Label.isNotApplicable()` method and a new `NotApplicableLabel` class.
- Updated the `TestVector` and `TestVectorBuilder` classes to understand not-applicable score values.
- Updated verifiers to expect not-applicable values
- Updated the `AverageCompositeScore` and `WeightedCompositeScore` classes to take into account not-applicable score values.
- Updated the `TuningWithCMAES` class to expect not-applicable values.
- Updated the `PrettyPrinter` to expect not-applicable values.
- Updated tests.

This closes #97

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
@@ -1,5 +1,7 @@
 package com.sap.sgs.phosphor.fosstars.model.qa;
 
+import com.sap.sgs.phosphor.fosstars.model.qa.TestVectorResult.Status;
+import com.sap.sgs.phosphor.fosstars.model.value.ScoreValue;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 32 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
```diff
@@ -62,4 +64,36 @@ public void verify() throws VerificationFailedException {
...
+   */
+  protected static TestVectorResult testResultFor(
+      TestVector vector, ScoreValue scoreValue, int index) {
+
+    if (vector.notApplicableScore() && scoreValue.isNotApplicable()) {
+      return new TestVectorResult(vector, index++, scoreValue,
+          Status.PASSED, "Ok, score is N/A as expected");
+    }
+
+    if (vector.notApplicableScore() && !scoreValue.isNotApplicable()) {
```
Sachpat, line 84: What about the case where vector is applicable and scoreValue is not applicable?

artem-smotrakov: I thought that `containsExpected()` handles that but in fact it doesn't. I'll fix it.

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/NotApplicableLabel.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  @Override
+  @JsonIgnore
+  public final boolean isNotApplicable() {
+    return true;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
Sachpat, line 24: Do we really need equals and hashcode for this class?

artem-smotrakov: I think so. With the current implementation of this class, all instances should be equal. `Object.equals()` doesn't guarantees it.

src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractVerifier.java
artem-smotrakov: That's a very good catch @Sachpat . Now it should be fixed. I also improved a test for the `ScoreVerifier`.

artem-smotrakov:

Issue #96, opened 2020-04-01
Introduce a feature for NoHTTP tool and include it to the security testing score


artem-smotrakov:
[nohttp](https://github.com/spring-io/nohttp) is a tool which helps to make sure that no HTTP is used in a project. Let's introduce a new boolean feature which shows if a project uses the tool.

The new feature should contribute to the security testing score. To include the new feature to the score, a new intermediate sub-score may be added. The sub-score should return the max score if the feature is true, and the min score otherwise.

Here is a list of things to do:

- [x] Implement the new feature.
- [x] Add a data provider which gathers the new feature.
- [ ] Update the score to use the new feature.
- [x] Add unit tests.
- [ ] Update the test vectors.
- [ ] Make sure that the demo tool works fine.

artem-smotrakov:
If a new special value "Not Applicable" is introduced (see #97 ), then the score can also use features which indicate that a project uses Maven or Gradle. If on, the score can return a "Not Applicable" value.

PR #130, opened 2020-04-21
Introduced a feature for NoHTTP tool

artem-smotrakov:

- Added a new feature `OssFeatures.USES_NOHTTP`.
- Added a new data provider `UsesNoHttpTool` to gather the new feature.
- Added `UsesNoHttpToolTest`.

This is a part of #96


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * @param repository The project's repository.
+   * @return True if the project uses the plugin, false otherwise.
+   */
+  private boolean checkMaven(GHRepository repository) throws IOException {
+    GHContent content;
+    try {
+      content = repository.getFileContent("pom.xml");
+    } catch (GHFileNotFoundException e) {
+      return false;
```
sourabhsparkala, line 64: An logger.warn() would be good here?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * @param repository The project's repository.
+   * @return True if the project uses the plugin, false otherwise.
+   */
+  private boolean checkGradle(GHRepository repository) throws IOException {
+    GHContent content;
+    try {
+      content = repository.getFileContent("build.gradle");
+    } catch (GHFileNotFoundException e) {
+      return false;
```
sourabhsparkala, line 95: logger here.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    if (content == null || !content.isFile()) {
+      return false;
+    }
+
+    try (BufferedReader reader = new BufferedReader(new InputStreamReader(content.read()))) {
+      String line;
+      while ((line = reader.readLine()) != null) {
+        if (line.trim().startsWith("id \"io.spring.nohttp\"")) {
```
sourabhsparkala, line 105: Can the string be a final static?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
artem-smotrakov: I am not sure about that. If there is no `pom.xml` file, then it is okay because a project may use other build system.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
artem-smotrakov: I am not sure it's necessary. Please see above.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpTool.java
artem-smotrakov: Yes, it can. Since it's used only in one place, let's keep it as-is. I think it makes it a bit easier to read code since you don't have to jump to the definition of the `final static`.

artem-smotrakov:

PR #135, opened 2020-04-22
Move GitHubDataFetcher to the com.sap.sgs.phosphor.fosstars.data.github package

### LINKED ISSUE #134: Move GitHubDataFetcher to the com.sap.sgs.phosphor.fosstars.data.github package
The `GitHubDataFetcher` fetches data from GitHub and mainly used by the classes from the `com.sap.sgs.phosphor.fosstars.data.github`. Let's move it to there.

artem-smotrakov:
The `GitHubDataFetcher` fetches data from GitHub and mainly used by the classes from the `com.sap.sgs.phosphor.fosstars.data.github`. Let's move it to there.

This closes #134

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/AbstractGitHubDataProvider.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/AbstractGitHubDataProvider.java
@@ -1,7 +1,6 @@
 package com.sap.sgs.phosphor.fosstars.data.github;
 
 import com.sap.sgs.phosphor.fosstars.data.AbstractDataProvider;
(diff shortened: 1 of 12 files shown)


sourabhsparkala:
But this is a DataFetcher right. Like GithubProjectFinder

artem-smotrakov:
> @artem-smotrakov
> But this is a DataFetcher right. Like GithubProjectFinder. So shouldn't this be in tool.github?

`GithubProjectFinder` is used only by the demo tool. It browses orgs and looks for projects. It doesn't really fetch data about specific projects.

The data fetcher help to fetch actual data. The class is used directly by multiple data providers. It's not used directly by the demo tool. That's why I think it should be moved to the data providers.

PR #136, opened 2020-04-22
Improve cache in data providers

### LINKED ISSUE #128: Re-structure the caching mechanism
We have two different caching interfaces

- `ValueCache`
- `DataCache`

It would be better to combine this by extending one interface. Also, it would be better to have a single storing mechanism (either json files or LRUMap or an alternative).

The changes done in the interfaces above should reflect in the implemented classes as well.

artem-smotrakov:

- Renamed `DataCache` to `Cache`.
- Renamed `NoCache` to `NoValueCache`.
- Moved `GitHubDataCache` to the `com.sap.sgs.phosphor.fosstars.data.github` package.
- `ValueCache` now extends `Cache`.
- Added new methods to `StandardValueCache`, `ValueSet`, `ValueHashSet`
  and `GitHubProjectValueCache`.
- `GitHubCachingDataProvider.fetchValuesFor()` method now returns a `ValueSet`.
- Added extra tests for caches.

This closes #128


src/test/java/com/sap/sgs/phosphor/fosstars/data/StandardValueCacheTest.java
```diff
@@ -36,9 +43,63 @@ public void store() throws IOException {
       assertTrue(content.length > 0);
 
       StandardValueCache anotherCache = StandardValueCache.load(filename);
-      assertEquals(cache, anotherCache);
+      assertEquals(cache, cache);
```
sourabhsparkala, line 46: assertEquals(cache, anotherCache);

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -2,24 +2,184 @@
...
+    assertEquals(1, fetcher.commitsCache().size());
+    assertEquals(commits.size(), fetchedCommits.size());
+
+    // fill out the cache for commits
+    int i = 0;
+    while (fetcher.commitsCache().size() < fetcher.commitsCache().maxSize()) {
+      project = new GitHubProject(
+          String.format("org%d", i), String.format("project%d", i));
+      fetcher.commitsFor(project, github);
+      i++;
```
sourabhsparkala, line 131: If this was ++i
then the below i++ is not required.

src/test/java/com/sap/sgs/phosphor/fosstars/data/StandardValueCacheTest.java
artem-smotrakov: This line checks that `cache` equals to itself. The next line checks if `cache` equals to `anotherCache`.

artem-smotrakov:

PR #137, opened 2020-04-23
 Added new NoHttpToolScore to ProjectSecurityTestingScore

### LINKED ISSUE #96: Introduce a feature for NoHTTP tool and include it to the security testing score

artem-smotrakov:

- Added a new `NoHttpToolScore` based on `USES_NOHTTP` and `PACKAGE_MANAGERS` features.
- Added `NoHttpToolScore` to `ProjectSecurityTestingScore`.

This closes #96


artem-smotrakov:

PR #138, opened 2020-04-24
Look for unpatched vulnerabilities in NVD

### LINKED ISSUE #2: NVDTest should not depend on the NVD
`NVDTest` is currently disabled since it requires downloading data from the NVD. The test needs to be updated to test only parsing data. It should not try to download data from the NVD.

### LINKED ISSUE #107: Look for unpatched vulnerabilities in NVD
Entries in NVD contains CPEs that shows which products and versions are affected. If a vulnerability doesn't have a fix, its CPEs may contain stars in specific components, for example `cpe:2.3:a:odata4j_project:odata4j:0.7:*:*:*:*:*:*:*`. Let's try to figure out if this info can be used to identify unpatched vulnerabilities. The issues mentioned in #106 may be used as a reference. If it works, then the `VulnerabilitiesFromNvd` data provider may be updated to search for unpatched issues.

artem-smotrakov:

- Updated the `UnpatchedVulnerabilities` data provider to look for
  unpatched vulnerabilities in NVD.
- Added `VendorDataMatcher` (based on `ExactMatcher` which was removed).
- Added `Vulnerability.Builder` class.
- Improved the `NVD` class.
- Added additional fields to the `CpeMatch` class.
- Updated the demo tool to provide an instance of the `NVD` class
  to the rating calculators.
- Added `UnpatchedVulnerabilitiesTest`.
- Updated `NVDTest`, this fixes #2.
- Updated tests for calculators and the demo tool.

This fixes #107


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
```diff
@@ -44,6 +60,71 @@ protected Feature supportedFeature() {
...
+   * Looks for unpatched vulnerabilities in NVD for a specified project.
+   *
+   * @param project The project.
+   * @return Unpatched vulnerabilities.
+   */
+  Vulnerabilities vulnerabilitiesFromNvdFor(GitHubProject project) {
+    Vulnerabilities vulnerabilities = new Vulnerabilities();
+    for (NvdEntry entry : nvd.search(VendorDataMatcher.with(project))) {
+      if (isUnpatched(entry)) {
+        vulnerabilities.add(
```
Sachpat, line 81: At this point, logically speaking they should also be added into knownUnpatchedVulnerabilities right? So that they will be taken from the same for the next search? Also shouldn't we filter those vulnerabilities which are already part of knownUnpatchedVulnerabilities?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
```diff
@@ -44,6 +60,71 @@ protected Feature supportedFeature() {
...
+    if (entry.getConfigurations() == null || entry.getConfigurations().getNodes() == null) {
+      return false;
+    }
+
+    for (Node node : entry.getConfigurations().getNodes()) {
+      if (node == null || node.getCpeMatch() == null) {
+        continue;
+      }
+
+      for (CpeMatch cpeMatch : node.getCpeMatch()) {
```
Sachpat, line 108: shouldn't the method name be node.getCpeMatches* ?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+    return match(project.name(), productData.getProductName());
+  }
+
+  /**
+   * Returns true if one of two strings includes the other one.
+   */
+  private static boolean match(String one, String two) {
+    one = one.toLowerCase().trim();
+    two = two.toLowerCase().trim();
+    return one.contains(two) || two.contains(one);
```
Sachpat, line 129: Does this mean that Ehcache and cache will be considered as a match? Is that what we really want?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
artem-smotrakov: This class and methods were generated from the JSON schema. We try to to touch them much although this pull request adds a couple of new methods to the class. If we decide to rename methods, then we should do it for the all generated classes. This is out of scope here.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
artem-smotrakov: > At this point, logically speaking they should also be added into knownUnpatchedVulnerabilities right? So that they will be taken from the same for the next search? 

No, `knownUnpatchedVulnerabilities` reads vulnerabilities from a file. The file is updated manually.

> Also shouldn't we filter those vulnerabilities which are already part of knownUnpatchedVulnerabilities?

No, the `Vulnerabilities` class merges vulnerabilities by IDs (in the `fetchValueFor()` method).

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: What do you think is a better way?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/InfoAboutVulnerabilities.java
```diff
@@ -37,11 +39,12 @@
    *
    * @param github An interface to the GitHub API.
    */
-  public InfoAboutVulnerabilities(GitHub github) throws IOException {
+  public InfoAboutVulnerabilities(GitHub github, NVD nvd) throws IOException {
```
sourabhsparkala, line 42: Maybe NVD should be initialized in this class, rather than being passed as an argument? What do you think?

artem-smotrakov: I think that the class needs to be configured with an instance of `NVD`.

artem-smotrakov: `NVD.download()` needs to be called before using the instance. This cause the class to download the data from NVD. We need to make sure that this is done once. Now, it's done by the demo tool, and it passes an initialized `NVD` instance to everyone who needs it.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+   * @return The new matcher.
+   */
+  public static VendorDataMatcher with(GitHubProject project) {
+    return new VendorDataMatcher(project);
+  }
+
+  /**
+   * A project to be checked.
+   */
+  private final GitHubProject project;
```
sourabhsparkala, line 40: Please move this line above the constructor.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+    if (cve == null) {
+      LOGGER.warn("No CVE in NVD entry");
+      return false;
+    }
+
+    CVEDataMeta meta = cve.getCveDataMeta();
+    if (meta == null) {
+      LOGGER.warn("No metadata in NVD entry");
+      return false;
+    }
```
sourabhsparkala, line 65: Maybe a line break is needed here.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: Okay.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+
+    Vendor cveVendor = affects.getVendor();
+    if (cveVendor == null) {
+      LOGGER.warn("No vendor in NVD entry for {}", cveId);
+      return false;
+    }
+
+    for (VendorData vendorData : cveVendor.getVendorData()) {
+      if (match(vendorData)) {
```
sourabhsparkala, line 85: I think you can just return this if predicate directly. It can return false also.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Checks if vendor data from an NVD entry matches to the project's owner and name.
+   *
+   * @param vendorData The vendor data.
+   * @return True if the vendor data matches the project, false otherwise.
+   */
+  private boolean match(VendorData vendorData) {
+    if (!match(project.organization().name(), vendorData.getVendorName())) {
```
sourabhsparkala, line 100: same here return the predicate

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
```diff
@@ -0,0 +1,131 @@
...
+      return false;
+    }
+
+    for (ProductData productData : vendorData.getProduct().getProductData()) {
+      if (match(productData)) {
```
sourabhsparkala, line 105: same here return the predicate

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -1,26 +1,55 @@
...
     PATCHED, UNPATCHED, UNKNOWN
   }
 
+  /**
+   * This means that a vulnerability has no description.
+   */
   public static final String NO_DESCRIPTION = null;
+
+  /**
+   * This means that a vulnerability has not refenrences.
```
sourabhsparkala, line 41: references

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -1,26 +1,55 @@
...
+   */
   public static final List<Reference> NO_REFERENCES = Collections.emptyList();
+
+  /**
+   * This means that it's unknown when a vulnerability was introduced.
```
sourabhsparkala, line 46: it is?

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -1,26 +1,55 @@
...
+   */
   public static final Date UNKNOWN_INTRODUCED_DATE = null;
+
+  /**
+   * This means that it's unknown when a vulnerability was fixed.
```
sourabhsparkala, line 51: it is?

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -208,17 +237,136 @@ public int hashCode() {
...
+  /**
+   * Returns a description of the vulnerability.
+   * This method is here to make Jackson happy.
+   */
   private String getDescription() {
     return description;
   }
+
+  /**
+   * This is builder that builds instances of {@link Vulnerability}.
```
sourabhsparkala, line 265: This is a builder

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
Sachpat: why don't we simply compare or do equals?

artem-smotrakov: Because it's not going to work. If you look at the existing CPE, you can notice that they don't match with organizations and project names on GitHub.

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/NVD.java
```diff
@@ -91,36 +94,84 @@ String location() {
...
   }
 
   /**
-   * Looks for NVD entries for a specific vendor and product.
+   * Creates an input stream for a file.
+   *
+   * @param file The file.
+   * @return The input stream.
+   */
+  private static InputStream open(Path file) {
```
sourabhsparkala, line 124: Do you think we need some Utils class to handle such methods
- Read file
- Do a HTTP call

A clean-up task?

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
artem-smotrakov: What is wrong with "it's"?

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/NVD.java
```diff
@@ -135,14 +186,11 @@ String location() {
         }
 
         for (NvdEntry entry : MAPPER.readValue(parser, NvdEntry[].class)) {
-          for (Matcher matcher : MATCHERS) {
-            if (matcher.match(entry, vendor, product)) {
-              nvdEntries.add(entry);
-            }
-          }
+          nvdEntries.put(entry.getCve().getCveDataMeta().getID(), entry);
```
sourabhsparkala, line 189: What if getCveDataMeta() is null?

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
artem-smotrakov: What is wrong with "it's"?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
Sachpat: I didn't know JSON schema was that dumb!

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
artem-smotrakov: Sure.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Vulnerability.java
artem-smotrakov: Sure.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: Sure.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: I am not sure if I understand what you mean.

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/data/CpeMatch.java
```diff
@@ -45,6 +53,16 @@ public void setCpe23Uri(String cpe23Uri) {
     this.cpe23Uri = cpe23Uri;
   }
 
+  @JsonProperty("versionEndExcluding")
+  public String getVersionEndExcluding() {
```
sourabhsparkala, line 57: Did I miss something here. Where are these being `set`?
- versionEndExcluding
- versionEndIncluding

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/NVD.java
artem-smotrakov: I am not sure about it.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
```diff
@@ -43,9 +49,13 @@
    * Initializes a new calculator.
    *
    * @param github An interface to GitHub.
+   * @param nvd An interface to NVD.
    */
-  AbstractRatingCalculator(GitHub github) {
-    this.github = Objects.requireNonNull(github, "Oh no! An interface to GitHub can't be null!");
+  AbstractRatingCalculator(GitHub github, NVD nvd) {
+    Objects.requireNonNull(github, "Oh no! An interface to GitHub can't be null!");
```
sourabhsparkala, line 55: can be written like this right?
```
this.github = Objects.requireNonNull(github, "Oh no! An interface to GitHub can't be null!");
this.nvd = Objects.requireNonNull(nvd, "Oh no! An interface to NVD can't be null!");
```

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/NVD.java
artem-smotrakov: That would mean that an entry in NVD doesn't have a CVE id. Currently, NVD doesn't have such entries. I can add null-checks for completeness.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
sourabhsparkala: return `match(vendorData);`

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UnpatchedVulnerabilities.java
artem-smotrakov: Sure. You can tell the NVD team about it.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: There may be multiple `VendorData`. If first one doesn't match, it has to check the next one.

src/main/java/com/sap/sgs/phosphor/fosstars/nvd/data/CpeMatch.java
artem-smotrakov: They are set during deserialization of JSON data from NVD.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/github/AbstractRatingCalculator.java
artem-smotrakov: Right. Your code may leave a partly-initialized instance. That's not a bid deal though. But I would like to check all parameters before using them. Just trying to build a good habit.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/VendorDataMatcher.java
artem-smotrakov: Neither `equals()` nor the current approach work well. I think we need to write a smarter matcher. It may use some fuzzy search and take into account other fields. Then, we'll need to verify how it works on real projects. Let's do it later.

artem-smotrakov: #145

artem-smotrakov:

PR #139, opened 2020-04-24
Make UsesGithubForDevelopment data provider smarter

### LINKED ISSUE #122: Make UsesGithubForDevelopment data provider smarter
An issue in #121 

We use [getMirrorUrl()](https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHRepository.html#getMirrorUrl()) to check if the given repository is an original repository or a mirror repository. 

```
if (repository.getMirrorUrl() == null || repository.getMirrorUrl().isEmpty()){
    System.out.print("Repository is a original repository");
}
System.out.print("Repository is a mirror repository");
```
But for mirror repos like [apache/tomee](https://github.com/apache/tomee). The `mirror_url` is `null`. Hence, the above check will fail. This issue can be seen in many mirror repos of [apache](https://github.com/apache). 

For reference: The `mirror_url` flag is `true` for [coreutils/coreutils](https://github.com/coreutils/coreutils)

Hence, find an alternative to confirm the verification if a repository is a mirror or an original repository.

artem-smotrakov:
Here is a list of updates:

- Updated the `UsesGithubForDevelopment` data provider to consider
  multiple facts about a project.
- Updated `UsesGithubForDevelopmentTest`.

This closes #122

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
@@ -7,19 +7,75 @@ import com.sap.sgs.phosphor.fosstars.model.Value;
 import com.sap.sgs.phosphor.fosstars.model.value.BooleanValue;
 import com.sap.sgs.phosphor.fosstars.tool.github.GitHubProject;
 import java.io.IOException;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
 import org.apache.commons.lang3.StringUtils;
+import org.kohsuke.github.GHCommit;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.github.GitHub;
 
 /**
- * This data provider checks if a project uses GitHub as a main development platform. The check
- * would be based on mirror url is null or empty. This will ensure whether the underlying project is
- * a mirror project or not.
+ * This data provider tries to figure out if a project uses GitHub for development.
  */
 public class UsesGithubForDevelopment extends CachedSingleFeatureGitHubDataProvider {
 
   /**
-   * Initializes a data provider.
+   * This threshold shows how many checks have to be passed to say that a project uses GitHub
+   * for development.
+   */
+  static final double CONFIDENCE_THRESHOLD = 0.7;
+
+  /**
+   * This threshold shows how many commits have to point to existing GitHub user names
+   * to say that commits are mostly done by users on GitHub.
+   */
+  private static final double RESOLVED_USERS_THRESHOLD = 0.9;
+
+  private static final List<Check> CHECKS = Arrays.asList(
+
+      // check if the repository doesn't have an explicit link to the original repository
+      (repository, commits) -> StringUtils.isEmpty(repository.getMirrorUrl()),
+
+      // check if the description mentions "mirror"
+      (repository, commits) -> !repository.getDescription().toLowerCase().contains("mirror"),
+
+      // if GitHub issues are enabled, then it's likely that the project uses GitHub
+      (repository, commits) -> repository.hasIssues(),
+
+      // if GitHub pages are enabled, then it's likely that the project uses GitHub
+      (repository, commits) -> repository.hasPages(),
+
+      // if GitHub Wiki is enabled, then it's likely that the project uses GitHub
+      (repository, commits) -> repository.hasWiki(),
+
+      // if one of the three options for pull requests is enabled, then it looks like
+      // that the project uses pull requests
+      (repository, commits) ->  repository.isAllowMergeCommit()
+          || repository.isAllowRebaseMerge() || repository.isAllowSquashMerge(),
+
+      // check if repository is not archived
+      (repository, commits) ->  !repository.isArchived(),
+
+      // check if the project doesn't have an explicit link to an SVN repository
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -39,31 +56,100 @@ protected Value fetchValueFor(GitHubProject project) throws IOException {
...
+   * @param commit The commit to be checked.
+   * @return True if user names from the commit can be resolved to existing GitHub users,
+   *         false otherwise.
    */
-  private boolean isMirror(GitHubProject project) throws IOException {
-    GHRepository repository = gitHubDataFetcher().repositoryFor(project, github);
-    return StringUtils.isNotEmpty(repository.getMirrorUrl());
+  private static boolean canResolveUsersFor(GHCommit commit) {
+    try {
+      commit.getCommitter();
```
sourabhsparkala, line 147: I think we need a null check here and in the next line. What are we checking here is not clear to me.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -39,31 +56,100 @@ protected Value fetchValueFor(GitHubProject project) throws IOException {
...
+  boolean usesGitHubForDevelopment(
+      GHRepository repository, List<GHCommit> commits, double threshold) {
+
+    List<Predicate<GHRepository>> checks = Arrays.asList(
+
+        // check if the repository doesn't have an explicit link to the original repository
+        r -> StringUtils.isEmpty(r.getMirrorUrl()),
+
+        // check if the description mentions "mirror"
+        r -> !r.getDescription().toLowerCase().contains("mirror"),
```
sourabhsparkala, line 98: Projects like 
- https://github.com/HannahMitt/HomeMirror
- https://github.com/MichMich/MagicMirror
would send the wrong information. But I guess the threshold is making sure that we do not look for a hard `true` to everything?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -39,31 +56,100 @@ protected Value fetchValueFor(GitHubProject project) throws IOException {
...
    * 
-   * @param project of type {@link GitHubProject}.
-   * @return true if the project is a mirror project. Otherwise, false.
-   * @throws IOException may throw an exception during REST call to GitHub API.
+   * @param repository The project's repository.
+   * @param commits The project's commit history.
+   * @return True if it looks like that the project uses GitHub, false otherwise.
+   *
+   */
+  boolean usesGitHubForDevelopment(
```
sourabhsparkala, line 89: I think this method could be static. As this is stateless.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -39,31 +56,100 @@ protected Value fetchValueFor(GitHubProject project) throws IOException {
...
+      GHRepository repository, List<GHCommit> commits, double threshold) {
+
+    List<Predicate<GHRepository>> checks = Arrays.asList(
```
sourabhsparkala, line 92: This arraylist can be static as well.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
artem-smotrakov: If a project doesn't use GitHub as the main development repository, then user ids in the commit messages are not necessary point to GitHub user ids. The `getCommitter()` and `getAuthor()` methods throw an i/o exception if they can't find a GitHub user for a username from the commit message.

I'll add a null check although it I am not sure if the GitHub data fercher can return nulls in the commit list. Anyway, it's better to have a check.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
artem-smotrakov: Correct.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
artem-smotrakov: Right. That would also simplify the test.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
artem-smotrakov: Not really at the moment. The lambdas in the array use the `commits` parameter but the `Predicate` interface takes only one parameter and it is `GHRepository`. To overcome this, I need to introduce a new interface for a check. The new interface should take both repository and list of commits as a parameter. At first, I thought using the `Predicate` interface is better (and shorter). I'll introduce a new interface for checks.

artem-smotrakov:
> Minor comments.
> Everything else looks good

Thanks for the review @sourabhsparkala ! I've addressed the comments.

PR #166, opened 2020-05-04
GitHubProjectFinder should take into account number of stars

### LINKED ISSUE #165: GitHubProjectFinder should take into account number of stars
Let's update `GitHubProjectFinder` to take into account number of stars in projects. This setting should be optional in a config.

artem-smotrakov:
This update introduce a new parameter for `GitHubProjectFinder` to make it filter out projects with a low number of stars. That would allow to exclude unpopular projects from the markdown report.

This closes #165

### DIFF
--- a/docs/oss/security/apache.yml
+++ b/docs/oss/security/apache.yml
@@ -9,6 +9,7 @@ reports:
 finder:
   organizations:
     - name: apache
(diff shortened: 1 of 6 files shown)


artem-smotrakov:

PR #167, opened 2020-05-04
Update UsesOwaspDependencyCheck to support Gradle

### LINKED ISSUE #11: UsesOwaspDependencyCheckTest should not depend on GitHub API
The test fails when the GitHub API rate limit exceeded or some networking issues occur. That's why the test is currently disabled.

The test needs to be re-written to avoid connecting to the GitHub API.

### LINKED ISSUE #131: Update UsesOwaspDependencyCheck to support Gradle
Currently the `UsesOwaspDependencyCheck` data provider can only read Maven configuration. But OWASP Dependency Check plugin may be also used with Gradle. Let's update the provider to support Gradle:

https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyCheck.java#L98

artem-smotrakov:

- Updated `UsesOwaspDependencyCheck` to read `build.gradle` file
  and check if OWASP Dependency Check plugin is used.
- Updated `UsesOwaspDependencyCheckTest`.

This closes #131
This closes #11


artem-smotrakov:

PR #169, opened 2020-05-05
Updated GitHubDataFetcher to clone repositories

### LINKED ISSUE #168: GitHubDataFetcher should be able to clone repositories
A data provider may need to check many files in a repository. It may not be convenient and efficient if the provider uses the GitHub API. Instead, it may be better to clone the repository and check the files locally. Let's teach `GitHubDataFetcher` to clone repositories.

artem-smotrakov:

- Updated `GitHubDataFetcher` to be able to clone repositories. It doesn't load commits via GitHub API anymore. Instead, it returns a `LocalRepository` class that loads the commit history from a local repository.
- `GitHubDataFetcher` is not a singleton anymore. It should be configured with an instance of `GitHub` and a path where it can store cloned repositories.
- `GitHubDataFetcher` has a new `cleanup()` method that cleans up local repositores. The method should be called with a strategy that decides which repositories should be removed.
- Updated the existing data providers to use the new API.
- Updated tests.

This closes #168


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -32,96 +41,123 @@
...
+   * Returns the interface to the GitHub API.
    */
-  private List<GHCommit> commitsFor(GHRepository repository) throws IOException {
-    try {
-      return repository.listCommits().toList();
-    } catch (HttpException e) {
-      LOGGER.error(String.format("Could not fetch commits from %s", repository.getUrl()), e);
-      return Collections.emptyList();
+  public GitHub github() {
+    return github;
```
sourabhsparkala, line 124: Could this ever return a null?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -144,50 +180,135 @@ public GHRepository repositoryFor(GitHubProject project, GitHub github) throws I
...
+
+    Optional<Repository> repository = openRepository(info.path());
+    if (!repository.isPresent()) {
+      Files.deleteIfExists(info.path());
+      clone(project, info.path());
+      repository = openRepository(info.path());
     }
 
-    return subList;
+    if (!repository.isPresent()) {
```
sourabhsparkala, line 233: Is there a way to combine this and the above if condition. I understand the significance of checking it twice. But could something else be written?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/LocalRepository.java
```diff
@@ -0,0 +1,212 @@
...
+   */
+  @JsonIgnore
+  public Optional<Commit> firstCommit() throws IOException {
+    loadCommitsIfNecessary();
+
+    if (commits.isEmpty()) {
+      return Optional.empty();
+    }
+
+    return Optional.of(commits.get(commits.size() - 1));
```
sourabhsparkala, line 115: I am assuming that the commits are sorted in reverse order?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
```diff
@@ -27,85 +24,71 @@
...
+   * Check if a project uses signed commits.
    *
    * @param project The project.
-   * @return boolean True if the project uses verified signed commits. Otherwise, False
+   * @return boolean True if the project uses signed commits. Otherwise, False
    * @throws IOException caused during REST call to GitHub API.
    */  
   private boolean askGithub(GitHubProject project) throws IOException {
-    List<GHCommit> yearCommits = gitHubDataFetcher().commitsAfter(ONE_YEAR_AGO, project, github);
+    List<Commit> yearCommits = fetcher.localRepositoryFor(project).commitsAfter(ONE_YEAR);
```
sourabhsparkala, line 83: Now I think about it, maybe we should change the name `commitsAfter()` -> `commitsWithin()`

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: Nope. The field is final and there is a null-check in the constructor.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: Those two check, which look similar, in fact check a bit different things. The first check checks if an an existing repository can be loaded. If not, they we do a little cleanup and, clone a fresh repository, and then try to open it again. So, the second check checks if the fresh repository can be opened.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
artem-smotrakov: Let me find a better name for the constant instead.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: ok

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
sourabhsparkala: ok

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: I meant can we move the lines #L228 and #L229 inside the `openRepository()`. So that we do not have to check if `!repository.isPresent()` twice.

In the end if !repository.isPresent() is still true then throw the exception.

Does it make sense?

artem-smotrakov: The lines #L228 and #L229 should not be moved to the `openRepository()`. That would mean that we always want to clone a fresh repository. But we want to re-use existing ones. The second `!repository.isPresent()` could be moved inside the first `if` but I wanted to avoid nested ifs.

artem-smotrakov: This is a fallback logic. We first check if we have a valid repository. If not, we try to fetch a fresh one (fallback). If we still can't open the repository, then we break.

sourabhsparkala: ok

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSignedCommits.java
artem-smotrakov: Actually, that's a good point to add a new `commitsWithin()` method but it should take a `Duration`.

artem-smotrakov: Just added the method.

artem-smotrakov:

PR #172, opened 2020-05-08
Better parsing data from NVD

### LINKED ISSUE #171: Better parsing data from NVD
There are a number of issues with the current implementation of parsing data from NVD:

- It uses version 1.0 but it's better to migrate to 1.1
- It loads all the fields from the NVD entries. They are not really used but consume memory and time for parsing.
- The `UnpatchedVulnerabilities` data provider have many false-positives.

### LINKED ISSUE #175: Run CI/CD jobs on Linux, Window and Mac
Let's update the job in GitHub Action to run on Linux, Windows and Mac.

artem-smotrakov:
Here is a list of main updates.

- Switched to NVD schema 1.1.
- Removed unnecessary code from the class that hold data from NVD. That reduces memory footprint and should fix OOM on Windows.
- Excluded unused fields in the classes that hold data from NVD.
- Updated `NVD.parse()` method not to open all files in the same time.
- Updated `UnpatchedVulnerabilities` provider to consider only
  configurations for a specific projects. Also, a vulnerability is
  considered unpatched only if all related configurations don't
  contain a fix version. That fixes may false-positives.
- Updated data providers to use different features for vulnerabilities
  to avoid invalid cache hits.
- Renamed `VendorDataMatcher` to `NvdEntryMatcher`
  and updated it to take into account configurations.
- Updated tests.
- Updated the job in GitHub Actions to run on Linux, Windows and Mac. This closes #175.

This closes #171


src/main/java/com/sap/sgs/phosphor/fosstars/nvd/data/CVE.java
```diff
@@ -46,97 +55,149 @@
...
   }
 
-  @JsonAnyGetter
-  public Map<String, Object> getAdditionalProperties() {
-    return this.additionalProperties;
+  public enum DataType {
+
+    CVE("CVE");
+
+    private static final Map<String, com.sap.sgs.phosphor.fosstars.nvd.data.CVE.DataType> CONSTANTS
```
sourabhsparkala, line 127: Can this package be in the import section? Or is it better to change the name of the enum?

artem-smotrakov: Yeah, I'll make it shorter.

artem-smotrakov: Done.

PR #173, opened 2020-05-08
Implement a score for memory safety testing

### LINKED ISSUE #99: Implement new features for sanitizers
There are a number of sanitizers for code written in C/C++ such as:

- AddressSanitizer
- UndefinedBehaviorSanitizer
- MemoryLeakSanitizer
- ThreadSanitizer

The sanitizers allow catching multiple issues which may have security implications. For example, AddressSanitizer can catch memory corruptions.

Using sanitizers improves overall security testing for a project. Let's introduce new features which show if a project use those sanitizers.

- [x] Add new features.
- [x] Add data providers which fill out the new feature.

### LINKED ISSUE #100: Introduce a score for memory safety testing
A new score may be added to assess how well a project implements testing for memory safety issues (such as buffer overflows, use-after-free, etc). The new score may be based on the following features:

- usage of sanitizers, see #99 
- usage of programming languages, see #91 

The new score should contribute to the open-source security testing score. Since the score is going to make much sense only for projects which use C/C++, the score may consider returning a "Not Applicable" value (see #97) if the projects doesn't use C/C++.

At the moment, the score is supposed to be called something like "MemorySafetyTestingScore", however there may be a better name. In particular, the better name may say that the score takes care about threat safety, memory leaks, etc.

- [x] Add a new score
- [x] Add unit tests
- [x] Add test vectors

artem-smotrakov:

- Introduced features for AddressSanitizer, MemorySanitizer and UndefinedBehaviourSanitizer. This closes #99
- Added a new data provider that fills out the new features.
- Added a score for testing for memory-safety issues. The new score contributes to the security testing score. This closes #100

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSanitizers.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSanitizers.java
@@ -1,0 +1,155 @@
+package com.sap.sgs.phosphor.fosstars.data.github;
+
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_ADDRESS_SANITIZER;
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_MEMORY_SANITIZER;
+import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.USES_UNDEFINED_BEHAVIOR_SANITIZER;
+import static com.sap.sgs.phosphor.fosstars.model.other.Utils.setOf;
+
+import com.sap.sgs.phosphor.fosstars.model.Feature;
+import com.sap.sgs.phosphor.fosstars.model.ValueSet;
+import com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures;
+import com.sap.sgs.phosphor.fosstars.model.value.ValueHashSet;
+import com.sap.sgs.phosphor.fosstars.tool.github.GitHubProject;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * The data providers checks if a project uses sanitizers. It gathers the following features:
+ * <ul>
+ *   <li>{@link OssFeatures#USES_ADDRESS_SANITIZER}</li>
+ *   <li>{@link OssFeatures#USES_MEMORY_SANITIZER}</li>
+ *   <li>{@link OssFeatures#USES_UNDEFINED_BEHAVIOR_SANITIZER}</li>
+ * </ul>
+ */
+public class UsesSanitizers extends GitHubCachingDataProvider {
+
+  /**
+   * A compiler option that defines sanitizers.
+   */
+  private static final String SANITIZER_OPTION = "-fsanitize=";
+
+  /**
+   * A list of well-known file names of build configs.
+   */
+  private static final String[] BUILD_CONFIGS = {
+      ".travis.yml", "Configure", "CMakeLists.txt"
+  };
+
+  /**
+   * A list of well-known files extensions of build configs.
+   */
+  private static final String[] BUILD_CONFIG_SUFFIXES = {
+      ".ac", ".cmake"
+  };
+
+  /**
+   * Initializes a data provider.
+   *
+   * @param fetcher An interface to GitHub.
+   */
+  public UsesSanitizers(GitHubDataFetcher fetcher) {
+    super(fetcher);
+  }
+
+  @Override
+  protected Set<Feature> supportedFeatures() {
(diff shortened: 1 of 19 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSanitizers.java
```diff
@@ -0,0 +1,155 @@
...
+   * @return A list of lines that contain options that enable the sanitizers.
+   * @throws IOException If something went wrong.
+   */
+  static List<String> lookForSanitizers(String content) throws IOException {
+    List<String> lines = new ArrayList<>();
+
+    try (BufferedReader reader = new BufferedReader(new StringReader(content))) {
+      String line;
+      while ((line = reader.readLine()) != null) {
+        int start = line.indexOf(SANITIZER_OPTION);
```
Sachpat, line 142: why don't you simply split the line based on spaces or use StringTokenizer here?

artem-smotrakov: There may be multiple compiler options in a line. The method looks for value of `-fsanitize` option. It can contain a single value such as `-fsanitize=address` or multiple values such as `-fsanitize=address,memory`. The method should be able to find both cases. It may be possible to implement this with a tokenizer but I am not sure if it would be simpler.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/MemorySafetyTestingScore.java
```diff
@@ -0,0 +1,133 @@
...
+  public ScoreValue calculate(Value... values) {
+    Value<Languages> languages = find(LANGUAGES, values);
+    Value<Boolean> usesAddressSanitizer = find(USES_ADDRESS_SANITIZER, values);
+    Value<Boolean> usesMemorySanitizer = find(USES_MEMORY_SANITIZER, values);
+    Value<Boolean> usesUndefinedBehaviorSanitizer = find(USES_UNDEFINED_BEHAVIOR_SANITIZER, values);
+
+    ScoreValue scoreValue = scoreValue(MIN,
+        languages, usesAddressSanitizer, usesMemorySanitizer, usesUndefinedBehaviorSanitizer);
+
+    boolean applicable = languages.isUnknown() || languages.get().containsAnyOf(UNSAFE_LANGUAGES);
```
Sachpat, line 76: if languages.isUnknown() should it be still applicable?

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/MemorySafetyTestingScore.java
```diff
@@ -0,0 +1,133 @@
...
+    if (!applicable) {
+      return scoreValue.makeNotApplicable();
+    }
+
+    if (usesAddressSanitizer.orElse(Boolean.FALSE)) {
```
Sachpat, line 81: hmm this is some new syntax for me, do we really need .orElse(Boolean.FALSE)?

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/MemorySafetyTestingScore.java
artem-smotrakov: It's a good question. If the score is not give info about languages, then it currently prefers to be strict and assumes that the project uses an unsafe language. Of course, it's an assumption, and it may be wrong sometimes. If a project doesn't actually use unsafe languages, it can always provide the `LANGUAGES` feature, and then the score produces a better value.

Sachpat: Makes sense.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesSanitizers.java
Sachpat: Well it would look simpler if we use tokenizer or split the line based on " " because it took a while to understand for me to know why you are looking for an index of SANITIZER_OPTION. 

And you need not do this 
```
     int end = line.indexOf(" ", start);
          if (end < 0) {
            end = line.length();
          }

``` if you tokenize the line and search for the key in key=value pairs.

artem-smotrakov: Sure, we all are used to use different ways. One prefers tokenizing, another one prefers indexes. I won't argue about what is simpler. In the end, I think that neither your or mine solution works well because they don't expect a whitespace between `-fsanitize` options and they don't expect multiple `-fsanitize` options. I'll re-work the code to cover these cases.

Sachpat: Thanks 👯‍♂️ :D

artem-smotrakov: I updated the code. Instead of using `StringTokenizer`, it now uses `String.split()` since `StringTokenizer` is a legacy class. Unfortunately, the procedure now is more complex but I am not sure if we can avoid that since it now covers more cases.

PR #176, opened 2020-05-11
Add a score for FindSecBugs

### LINKED ISSUE #101: Introduce a feature for Find Security Bugs
[Find Security Bugs](https://find-sec-bugs.github.io/) is a free tool for static analysis for Java applications. The tool can be easily integrated into Maven or Gradle build.

Let's introduce a new boolean feature which shows if a project uses Find Security Bugs. The new feature can contribute to the security testing score. In order to integrate it to the score, we may need to introduce an intermediate sub-score which returns the max score if the feature is true, and the min value otherwise.

- [x] Implement the new feature.
- [x] Implement a data provider (or update the existing ones) which fills out the new feature.
- [x] Include the feature to the security testing score via an intermediate sub-score.
- [x] Update tests and test vectors.
- [x] Update the demo tool if necessary.

### LINKED ISSUE #174: Composite scores should include sub-scores even if they are N/A
Currently, composite scores don't include sub-score to the resulting score value if those sub-scores produced a N/A value. That doesn't seem to be correct. The score should include all values to the resulting value.

artem-smotrakov:

- Introduced new feature and score that show if a project uses FindSecBugs.
- Added a new data provider that gathers information about the new feature.
- Updated composite scores to include sub-scores even if they are N/A. This fixes #174.

This closes #101


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
```diff
@@ -87,11 +87,11 @@ public ScoreValue calculate(Value... values) {
     int numberOfSubScores = 0;
     for (Score subScore : subScores) {
       ScoreValue subScoreValue = calculateIfNecessary(subScore, valueSet);
+      scoreValue.usedValues(subScoreValue);
```
sourabhsparkala, line 90: if the subscoreValue is not applicable, should it be considered as used?

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScore.java
```diff
@@ -124,14 +124,14 @@ public final ScoreValue calculate(Value... values) {
     ScoreValue scoreValue = new ScoreValue(this);
     boolean allNotApplicable = true;
     for (Score subScore : subScores) {
-      Weight weight = weights.of(subScore).orElseThrow(IllegalStateException::new);
       ScoreValue subScoreValue = calculateIfNecessary(subScore, valueSet);
+      scoreValue.usedValues(subScoreValue);
```
sourabhsparkala, line 128: Here also the same concern. If the subscore value is not applicable. Should it be used?

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AverageCompositeScore.java
artem-smotrakov: Yes, please see #174. A score should contain info about all considered sub-scores even if they are N/A.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/WeightedCompositeScore.java
artem-smotrakov: Please see above.

artem-smotrakov:

PR #180, opened 2020-05-12
Improve vulnerability matching

### LINKED ISSUE #145: Improve vulnerability matching
The `NvdEntryMatcher` is very simple and not reliable. We need to find a better way to find vulnerabilities in NVD for projects on GitHub.

https://github.com/SAP/fosstars-rating-core/pull/138#discussion_r415720767

sourabhsparkala:
- Putting in more checks to do a fuzzy similarity check in NvdEntryMatcher

Fixes #145

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
@@ -12,7 +12,11 @@ import com.sap.sgs.phosphor.fosstars.nvd.data.ProductData;
 import com.sap.sgs.phosphor.fosstars.nvd.data.Vendor;
 import com.sap.sgs.phosphor.fosstars.nvd.data.VendorData;
 import com.sap.sgs.phosphor.fosstars.tool.github.GitHubProject;
+import java.util.List;
 import java.util.Objects;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.similarity.JaroWinklerSimilarity;
+import org.apache.commons.text.similarity.LongestCommonSubsequence;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -25,6 +29,21 @@ public class NvdEntryMatcher implements Matcher {
    * A logger.
    */
   private static final Logger LOGGER = LogManager.getLogger(NvdEntryMatcher.class);
+
+  /**
+   * The threshold for the project to be checked.
+   */
+  private static final double METRIC_THRESHOLD = 0.9;
+
+  /**
+   * Instance of {@link LongestCommonSubsequence}.
+   */
+  private static final LongestCommonSubsequence LCS = new LongestCommonSubsequence();
+
+  /**
+   * Instance of {@link JaroWinklerSimilarity}.
+   */
+  private static final JaroWinklerSimilarity JWS = new JaroWinklerSimilarity();
 
   /**
    * A project to be checked.
@@ -87,30 +106,14 @@ public class NvdEntryMatcher implements Matcher {
       return false;
     }
 
-    for (Node node : configurations.getNodes()) {
-      if (node.getCpeMatches() == null) {
-        continue;
-      }
-
-      for (CpeMatch cpeMatch : node.getCpeMatches()) {
-        if (cpeMatch == null) {
-          continue;
-        }
-
-        if (match(cpeMatch.getCpe22Uri()) || match(cpeMatch.getCpe23Uri())) {
-          return true;
-        }
-      }
-    }
-
-    return false;
+    return parseNodes(configurations.getNodes());
   }
 
   /**
    * Returns true if a string matches the project, false otherwise.
    */
   private boolean match(String string) {
-    return string != null && string.toLowerCase().contains(project.name());
+    return match(project.name(), string.trim());
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -97,7 +105,10 @@ private boolean match(Configurations configurations) {
           continue;
         }
 
-        if (match(cpeMatch.getCpe22Uri()) || match(cpeMatch.getCpe23Uri())) {
+        if (StringUtils.isNotEmpty(cpeMatch.getCpe22Uri())
+            && match(productFrom22Uri(cpeMatch.getCpe22Uri()))
+            || (StringUtils.isNotEmpty(cpeMatch.getCpe23Uri())
+                && match(productFrom23Uri(cpeMatch.getCpe23Uri())))) {
```
artem-smotrakov, line 111: Could you please extract this complex check to a separate method (or a couple of methods)? If possible, please try to simplify it.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -173,4 +184,40 @@ private static boolean match(String one, String two) {
...
+  private String productFrom23Uri(String cpe23Uri) {
+    return cpe23Uri.split(":")[4];
+  }
+
+  /**
+   * Returns product name from {@link CpeMatch#getCpe22Uri()}.
+   */
+  private String productFrom22Uri(String cpe22Uri) {
+    return cpe22Uri.split(":")[3];
+  }
```
artem-smotrakov, line 200: Maybe it ArrayIndexOutOfBoundsException won't happen here, but please make sure that those elements in the array exist before accessing them.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -173,4 +184,40 @@ private static boolean match(String one, String two) {
     two = two.toLowerCase().trim();
     return one.contains(two) || two.contains(one);
```
artem-smotrakov, line 185: Is this method still used?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -173,4 +184,40 @@ private static boolean match(String one, String two) {
...
+   */
+  private boolean metric(String term1, String term2) {
+    int point = 0;
+
+    if ((new JaroWinklerDistance()).apply(term1, term2) > METRIC_THRESHOLD) {
+      point += 1;
+    }
+
+    if ((new LongestCommonSubsequence()).apply(term1,
+        term2) >= (term1.length() * METRIC_THRESHOLD)) {
```
artem-smotrakov, line 217: Please don't break the line here in the list of parameters. Please try to simplify this condition (for example, you can define methods, variable with meaningful names).

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
sourabhsparkala: You are right.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
sourabhsparkala: I am thinking about how to better use this. Why it is compared with `contains` when comparing organization and vendor name? Was there a specific reason?

artem-smotrakov: No strong reasons.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -156,21 +174,111 @@ private boolean match(VendorData vendorData) {
...
+   * @param cpeMatch of type {@link CpeMatch}.
+   * @return True if one of the cpeXXUri contains the project's name. Otherwise False.
+   */
+  private boolean cpeUriCheck(CpeMatch cpeMatch) {
+    if (StringUtils.isNotEmpty(cpeMatch.getCpe23Uri())) {
+      return match(productFrom23Uri(cpeMatch.getCpe23Uri()));
+    } else if (StringUtils.isNotEmpty(cpeMatch.getCpe22Uri())) {
+      return match(productFrom22Uri(cpeMatch.getCpe22Uri()));
+    }
+    return false;
```
artem-smotrakov, line 252: How about this:

```java
return match(productFrom23Uri(cpeMatch.getCpe23Uri()))
    || match(productFrom22Uri(cpeMatch.getCpe22Uri()))
```

`stringAtIndex()` will then need to return an empty string if `content` is null.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -156,21 +174,111 @@ private boolean match(VendorData vendorData) {
...
+
+    if (jaroWinklerSimilarityCheck(one, two)) {
+      point += 1;
+    }
+
+    if (longestCommonSubsequenceCheck(one, two)) {
+      point += 1;
+    }
+
+    return point > 0 && point / 2 > METRIC_THRESHOLD;
```
artem-smotrakov, line 209: `point / 2` results to an integer, not a double. If `point` is 1, then the result will always be 0. If you'd like to use a double threshold here, you need to do `(double) point / 2`. Although I am not sure if we need a threshold here if we have only two checks. The current check may be replaced with `jaroWinklerSimilarityCheck(one, two) && longestCommonSubsequenceCheck(one, two)`.

artem-smotrakov: We also have a bug in `UsesGithubForDevelopment` that also devides two integers

https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java#L111

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -156,21 +159,133 @@ private boolean match(VendorData vendorData) {
...
+   * This is recursive method to parse through the list of nodes. Also its children nodes if
+   * present. To find a match to the project's name.
+   * 
+   * @param nodes List of type {@link Node}.
+   * @return True if there was a match. Otherwise False.
+   */
+  private boolean parseNodes(List<Node> nodes) {
+    for (Node node : nodes) {
+
+      if (node.getChildren() != null && !node.getChildren().isEmpty()
```
artem-smotrakov, line 270: Could you please simplify this by moving the null check to the beginning of the method? The, `isEmpty()` check is not necessary any more.

```java
private boolean parseNodes(List<Node> nodes) {
  if (nodes == null) {
      return false;
  }

  for (Node node : nodes) {

    if (parseNodes(node.getChildren()) {
      return true;
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -87,30 +106,14 @@ private boolean match(Configurations configurations) {
...
-    return false;
+    return parseNodes(configurations.getNodes());
   }
 
   /**
    * Returns true if a string matches the project, false otherwise.
    */
   private boolean match(String string) {
-    return string != null && string.toLowerCase().contains(project.name());
+    return match(project.name(), string.toLowerCase().trim());
```
Sachpat, line 116: Here we are anyway doing equalsIgnoreCase https://github.com/SAP/fosstars-rating-core/pull/180/files#diff-399975d627508494b618551e2122f721R180 then why do we need string.toLowerCase() again?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
```diff
@@ -156,21 +159,133 @@ private boolean match(VendorData vendorData) {
...
+   * present. To find a match to the project's name.
+   * 
+   * @param nodes List of type {@link Node}.
+   * @return True if there was a match. Otherwise False.
+   */
+  private boolean parseNodes(List<Node> nodes) {
+    for (Node node : nodes) {
+
+      if (node.getChildren() != null && !node.getChildren().isEmpty()
+          && parseNodes(node.getChildren())) {
```
artem-smotrakov, line 271: Is there a reason why it checks children and then CPE matches in the current node? I feel like many entries don't have children, do they?

src/test/java/com/sap/sgs/phosphor/fosstars/nvd/NVDTest.java
```diff
@@ -54,4 +54,60 @@ public void find() throws IOException {
...
+          nvd.search(NvdEntryMatcher.entriesFor(new GitHubProject("openssl", "openssl-book")));
+      assertNotNull(entries);
+      assertEquals(0, entries.size());
+
+      // The below test case tests for both cpe22Uri and cpe23Uri.
+      entries = nvd
+          .search(NvdEntryMatcher.entriesFor(new GitHubProject("FasterXML", "jackson-databind")));
+      assertNotNull(entries);
+      assertEquals(5, entries.size());
+
```
artem-smotrakov, line 104: Please remove this extra line.

src/test/java/com/sap/sgs/phosphor/fosstars/nvd/NVDTest.java
```diff
@@ -54,4 +54,60 @@ public void find() throws IOException {
...
+  public void matcher() throws IOException {
+    TestNVD nvd = new TestNVD();
+    try (InputStream content = getClass().getResourceAsStream("NVD_matcher.json")) {
+      nvd.add("file.json", content);
+      nvd.parse();
+
+      List<NvdEntry> entries = nvd.search(
+          NvdEntryMatcher.entriesFor(new GitHubProject("spring-projects", "spring-framework")));
+      assertNotNull(entries);
+
```
artem-smotrakov, line 68: Please remove this extra line.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
sourabhsparkala: Good idea. This would be better.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/NvdEntryMatcher.java
sourabhsparkala: It was just my preference. I wanted to do DPS. Or do you feel BFS is better approach? I think in this case BFS makes a lot of sense.

artem-smotrakov: BFS looks better for me but it's not a big deal here. I'll levave it up to you.

sourabhsparkala:
> We discussed that the matcher can check references. Did you find that it's not going to work or you'd prefer to have it as a separate improvement?
Yes, you are right. But this would not be applicable here as of now. It is not a reliable field yet. There were 2 questions which hindered this check inclusion:
- The reference link may not have a link with github/org/name. Then what is the threshold here to confirm `true` or `false`.
- Which check in this flow would give absolute output. And which checks should be used with this reference check to confirm against the threshold?

> CPEs contain a vendor. For jackson-databind, it's fasterxml. For Apache projects, it's apache. Would it make sense to update the mather to consider this? (not necessary in this update).

Yes, I couldn't use it, as there are cases where CPEUri may contain a different Vendor name. 
E.g: `spring-projects/spring-framework` -> `cpe:2.3:a:pivotal_software:spring_framework:*:*:*:*:*:*:*:*`

artem-smotrakov:
> > We discussed that the matcher can check references. Did you find that it's not going to work or you'd prefer to have it as a separate improvement?
> 
> Yes, you are right. But this would not be applicable here as of now. It is not a reliable field yet. There were 2 questions which hindered this check inclusion:
> 
> * The reference link may not have a link with github/org/name. Then what is the threshold here to confirm `true` or `false`.
> * Which check in this flow would give absolute output. And which checks should be used with this reference check to confirm against the threshold?
> 

Sure, those questions need to be still answered.

> > CPEs contain a vendor. For jackson-databind, it's fasterxml. For Apache projects, it's apache. Would it make sense to update the mather to consider this? (not necessary in this update).
> 
> Yes, I couldn't use it, as there are cases where CPEUri may contain a different Vendor name.
> E.g: `spring-projects/spring-framework` -> `cpe:2.3:a:pivotal_software:spring_framework:*:*:*:*:*:*:*:*`

Sure. This check would not always work. But if org name is equal to vendor, would not it give a bit more confidence that the issue belongs to the project?

Could you please open an issue to investigate the options above?

src/test/java/com/sap/sgs/phosphor/fosstars/nvd/NVDTest.java
sourabhsparkala: I have removed this. I am not sure why it is not getting updated.

sourabhsparkala:
> > > We discussed that the matcher can check references. Did you find that it's not going to work or you'd prefer to have it as a separate improvement?
> > 
> > 
> > Yes, you are right. But this would not be applicable here as of now. It is not a reliable field yet. There were 2 questions which hindered this check inclusion:
> > 
> > * The reference link may not have a link with github/org/name. Then what is the threshold here to confirm `true` or `false`.
> > * Which check in this flow would give absolute output. And which checks should be used with this reference check to confirm against the threshold?
> 
> Sure, those questions need to be still answered.
> 
> > > CPEs contain a vendor. For jackson-databind, it's fasterxml. For Apache projects, it's apache. Would it make sense to update the mather to consider this? (not necessary in this update).
> > 
> > 
> > Yes, I couldn't use it, as there are cases where CPEUri may contain a different Vendor name.
> > E.g: `spring-projects/spring-framework` -> `cpe:2.3:a:pivotal_software:spring_framework:*:*:*:*:*:*:*:*`
> 
> Sure. This check would not always work. But if org name is equal to vendor, would not it give a bit more confidence that the issue belongs to the project?
> 
> Could you please open an issue to investigate the options above?

Address in #186

sourabhsparkala:
Thanks @artem-smotrakov and @Sachpat for the review.

PR #182, opened 2020-05-12
Introduced a score for fuzzing

### LINKED ISSUE #98: Introduce a feature for OSS-Fuzz
[OSS-Fuzz](https://github.com/google/oss-fuzz) offers continuous fuzzing of open-source software.

Let's introduce a feature which shows if a project is included to OSS-Fuzz. The new feature may be included to a new score for fuzzing. In the beginning, the score can just use this single feature. It can return the max score value if the feature is true, and the min score value otherwise.

The score should be then included into the security testing score.

- [x] Add the new feature.
- [x] Implement a data provider which can automatically fill the new feature out.
- [ ] Add the new score and add it into the security testing score.
- [ ] Updates tests and test vectors.

### LINKED ISSUE #177: Remove unused feature for security reviews and fuzzing attempts
The features for security reviews and fuzzing attempts are currently not used. Furthermore, those feature can be at the moment filled out only manually. Let's remove them.

artem-smotrakov:
Here is a list of main updates

- Added a data provider that gathers the `FUZZED_IN_OSS_FUZZ` feature.
- Added a new `FuzzingScore`.
- Included the new score to the security testing score. This closes #98.
- Remove unused feature for security reviews and fuzzing attempts. This closes #177.


artem-smotrakov:

PR #183, opened 2020-05-12
VulnerabilityLifetimeScore should not break when dates are incorrect

### LINKED ISSUE #178: IllegalArgumentException when calculating a rating for Apache POI
The OSS security rating calculation fails with the following exception for https://github.com/apache/poi:

```
[+] Here is what we know about the project:
[+]    If an open-source project belongs to Eclipse Foundation: false
[+]    If a project uses Dependabot: false
[+]    If a project uses GitHub as the main development platform: false
[+]    If an open-source project uses AddressSanitizer: false
[+]    If an open-source project uses FindSecBugs: false
[+]    Info about vulnerabilities in open-source project: 960 vulnerabilities
[+]    Number of stars for a GitHub repository: 952
[+]    Number of contributors in the last three months: 7
[+]    Number of commits in the last three months: 136
[+]    If an open-source project belongs to Apache Foundation: true
[+]    If a project uses signed commits: false
[+]    If an open-source project is regularly scanned for vulnerable dependencies: unknown
[+]    If an open-source project has a security team: unknown
[+]    If an open-source project uses MemorySanitizer: false
[+]    If a project uses nohttp tool: false
[+]    If a project uses LGTM: true
[+]    Number of watchers for a GitHub repository: 79
[+]    A set of programming languages: RUBY, JAVA, OTHER, SCALA, CPP
[+]    If an open-source project has a security policy: false
[+]    If an open-source project is included to OSS-Fuzz project: false
[+]    If an open-source project uses UndefinedBehaviorSanitizer: false
[+]    When a project started: 2002-01-31T03:22:28.000+0100
[+]    A set of package managers: GRADLE
[+]    The worst LGTM grade of a project: A
[+]    If an open-source project is supported by a company: false
[!] Oh no! Could not calculate a rating for https://github.com/apache/poi
[!] java.lang.IllegalArgumentException: Looks like the vulnerability had been fixed even before it was introduced! How is that possible?
```

artem-smotrakov:
This is a quick fix for #178 that updates the `VulnerabilityLifetimeScore` to ignore vulnerabilities with incorrect fixed and introduced dates. Most likely the score will be soon disabled since it's not reliable enough.

This closes #178

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
@@ -99,8 +99,10 @@ public class VulnerabilityLifetimeScore extends FeatureBasedScore {
       Instant fixed = entry.fixed().orElse(now).toInstant();
 
       if (fixed.isBefore(introduced)) {
-        throw new IllegalArgumentException("Looks like the vulnerability had been fixed "
-            + "even before it was introduced! How is that possible?");
+        logger.warn("Looks like the vulnerability {} had been fixed ({}) "
+                + "even before it was introduced ({})! How is that possible? Let's skip it.",
+            entry.id(), fixed, introduced);
+        continue;
       }
 
       CVSS cvss = entry.cvss();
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
```diff
@@ -21,6 +21,8 @@
 import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
```
sourabhsparkala, line 25: Are these extra imports?

artem-smotrakov: Yup. I'll remove them. Thanks!

Issue #188, opened 2020-05-13
Relative thresholds for labels


artem-smotrakov:
Currently, the security rating use hardcoded thresholds when it assigns a label for a project:

```java
  protected SecurityLabel label(double score) {
    Score.check(score);

    if (score < 5.0) {
      return SecurityLabel.BAD;
    }

    if (score < 8.0) {
      return SecurityLabel.MODERATE;
    }

    return SecurityLabel.GOOD;
  }
```

That may introduce some amount of bias. The person, who defines the thresholds, may be too strict or vice versa. It may be better to define threshold based on data about well-known open-source projects.

Issue #184, opened 2020-05-12
UsesGithubForDevelopment should not divide integers


artem-smotrakov:
https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java#L111

The data provider does `points / CHECKS.size()` that results to an integer, and then compares it to a double. The result of the division should be double.

Issue #189, opened 2020-05-13
PoC: Explore NVD REST API by NIST


sourabhsparkala:
Check if we can replace our already existing interface to `NVD`.

At present we download the [JSON feeds](https://nvd.nist.gov/vuln/data-feeds#JSON_FEED). To parse through the list of vulnerabilities and find the ones associated to the `GitHub` project.

Instead why not use the `REST API` available by `NVD NIST`. More information can be found [here](https://nvd.nist.gov/General/News/New-NVD-CVE-CPE-API-and-SOAP-Retirement).

E.g: [jackson-databind](https://services.nvd.nist.gov/rest/json/cves/1.0?cpeMatchString=cpe:2.3:::jackson-databind)

Things to do:
- Check if this `REST API` is more efficient than downloading the entire `JSON` feed.
- Find the Delta
  - Check which is more reliable form of getting the vulnerabilities. If the count of vulnerabilities `>` count of `JSON` feed vulnerabilities for a `GitHub` project.
  - If the vulnerabilities collected from `API` call are all `TRUE POSITIVE`.
- Effort required to implement the changes.
- Create a Follow-up Issue if the required effort is more.

PR #190, opened 2020-05-13
Analyse data and ratings for well-known open-source projects

### LINKED ISSUE #164: Analyse data and ratings for well-known open-source projects
`docs/oss/security/github_projects.json` contains data and ratings for well-known open-source projects. It would be good to analyse what make ratings good or bad. It may give us idea how the rating and data gathering may be improved. Let's create a Jupyter notebook with some statistics over the data and analyse it.

artem-smotrakov:
Please have a look at the stats and graphs for data and ratings for well-known open-source projects:

https://github.com/artem-smotrakov/fosstars-rating-core/blob/analyse-ratings/src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb

In the end of the notepad, there is a list of discovered issues and conclusions. I am copying them to here:

1. Vulnerability matching doesn't work well, there are a lot of false-positives (for example, see a number of issues for OpenSSL).
    TODO: Let's see if https://github.com/SAP/fosstars-rating-core/pull/180 improves it. If it doesn't, we'll need to disable scores that use info about vulnerabilities (`UnpatchedVulnerabilitiesScore` and `VulnerabilityLifetimeScore`). See #191
 2. Most of vulnerabilities are considered unpatched. Those are false-positives.
    TODO: Try to re-work `UnpatchedVulnerabilitiesScore`. If it doesn't work out, use only the list of unpatched issue. See #191
 3. All vulnerabilities don't have dates when they were introduced. Many vulnerabilities don't have dates when they fixed. The existing dates are estimated by the dates when the issues were published. Then, `VulnerabilityLifetimeScore` tries to estimate the dates. As a result, the score becomes very unreliable. See #191
    TODO: Try to re-work `VulnerabilityLifetimeScore`. If it doesn't work out, disable the score.
 4. Both `LgtmScore` and `FindSecBugsScore` assess how a project uses static analysis. It's not necessary for a project to use both (or even more tools). The score may be combined to a `StaticAnalysisScore`.
    TODO: Introduce `StaticAnalysisScore`. See #192
 5. The feature for a security team is always unknown. That looks like a bug in the corresponding data provider.
    TODO: Fix it. See #193
 6. The feature 'If a project uses GitHub as the main development platform' is almost always No.
    But many of the projects use GitHub for development. This looks like a bug in the data provider.
    TODO: Fix it. May be caused by https://github.com/SAP/fosstars-rating-core/issues/184. See #194
 7. None of the projects use sanitizers. That looks suspicious.
    TODO: Look into it. See #195
 8. `ProjectSecurityTestingScore` calculates an average of sub-scores. That may not be fair.
    TODO: Consider updating `ProjectSecurityTestingScore` to use weights. See #196

I'll open follow-up issues once this update is merged.

In the end of the notepad, you can also find a PoC of thresholds for labels that are based on the data about analysed projects. #188 has been created to implement it in `OssSecurityRating` if we find no issues with it.

Here is a list of main updates in this pull request:

- Added `src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb` that contains stats and graphs for data and ratings about well-known open-source projects.
- Updated the report for well-known open-source projects:
  - Added Spring projects.
  - Added FasterXML projects.
  - Added Netty projects.
  - Added OpenSSL, curl and a couple of projects from Google.
- Added a PoC for relative thresholds for labels.

This closes #164


Sachpat:
> TODO: Let's see if #180 improves it. If it doesn't, we'll need to disable scores that use info about vulnerabilities (UnpatchedVulnerabilitiesScore and VulnerabilityLifetimeScore).

Hmm, AFAIK, UnpatchedVulnerabilitiesScore is the most important score compared to others when it comes to Security Ratings right? We should somehow fix this. Disabling this score will not get us anywhere at least for Security Ratings I feel.

sourabhsparkala:
Good analysis report 👍

artem-smotrakov:
> Good analysis report

Thanks for the feedback @sourabhsparkala !

docs/oss/security/FasterXML/jackson-dataformats-text.md
```diff
@@ -89,21 +115,21 @@ Here is how the rating was calculated:
...
           Number of watchers for a GitHub repository:...15
 
       Sub-score:....Vulnerability lifetime
       Importance:...Low (weight 0.23 out of 1.00)
-      Value:........3.25  out of 10.00
+      Value:........10.00  out of 10.00
       Confidence:...10.00 out of 10.00
       Based on:...2 features:
-          Info about vulnerabilities:...3 vulnerabilities
+          Info about vulnerabilities:...0 vulnerabilities
```
sourabhsparkala, line 129: Is this the case of ``FALSE POSITIVE` cases getting fixed?

docs/oss/security/FasterXML/jackson-dataformat-xml.md
```diff
@@ -93,18 +118,18 @@ Here is how the rating was calculated:
...
+          Number of watchers for a GitHub repository:...32
 
       Sub-score:....Vulnerability lifetime
       Importance:...Low (weight 0.23 out of 1.00)
-      Value:........0.00  out of 10.00
+      Value:........6.25  out of 10.00
       Confidence:...10.00 out of 10.00
       Based on:...2 features:
-          Info about vulnerabilities:...5 vulnerabilities
+          Info about vulnerabilities:...2 vulnerabilities
```
sourabhsparkala, line 129: `FALSE POSITIVE` cases fixed here as well.

docs/oss/security/apache/activemq.md
```diff
@@ -94,18 +120,18 @@ Here is how the rating was calculated:
...
+          Number of stars for a GitHub repository:......1734
           Number of watchers for a GitHub repository:...203
 
       Sub-score:....Vulnerability lifetime
       Importance:...Low (weight 0.23 out of 1.00)
       Value:........0.00  out of 10.00
       Confidence:...10.00 out of 10.00
       Based on:...2 features:
-          Info about vulnerabilities:...27 vulnerabilities
+          Info about vulnerabilities:...32 vulnerabilities
```
sourabhsparkala, line 131: Could this be a `FALSE POSITIVE` case

[link](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&search_type=all&cpe_vendor=cpe%3A%2F%3Aapache&cpe_product=cpe%3A%2F%3A%3Aactivemq) returns only 29 results.

sourabhsparkala: I guess we are not there yet with the accuracy.

artem-smotrakov: Note that the report was created when https://github.com/SAP/fosstars-rating-core/pull/180 was not merged.

Sachpat:
> > > TODO: Let's see if #180 improves it. If it doesn't, we'll need to disable scores that use info about vulnerabilities (UnpatchedVulnerabilitiesScore and VulnerabilityLifetimeScore).
> > 
> > 
> > Hmm, AFAIK, UnpatchedVulnerabilitiesScore is the most important score compared to others when it comes to Security Ratings right? We should somehow fix this. Disabling this score will not get us anywhere at least for Security Ratings I feel.
> 
> Yes, `UnpatchedVulnerabilitiesScore` is an important one. The score for security testing is also important. Not sure which one is more important. I think I forgot to update the conclusion about `UnpatchedVulnerabilitiesScore`. It can still be enabled but we'll need to disable its part that looks for unpatched vulnerabilities in NVD. That part produces a lot of false-positives. The part, which uses a manually updated list of unpatched issues, still can be used.
> 
> At the moment I am out of ideas how to improve those scores. I think it they produce bad results, it's worth to disable them until we find a solution. Again, `UnpatchedVulnerabilitiesScore` will likely stay, but `VulnerabilityLifetimeScore` will likely be disabled.

Just an idea just based on my understanding, is it possible to borrow that logic of how Vulas colleagues identify which version fixes it or if there is a fixed version already available. So that out of the list of unpatched vulnerabilities, if for any vulnerability, the fixed version of the library is available, it's a false positive and we can remove it from the list.

docs/oss/security/FasterXML/jackson-dataformats-text.md
artem-smotrakov: I think it should be 2 issues

https://www.cvedetails.com/vulnerability-list/vendor_id-15866/product_id-61565/Fasterxml-Jackson-dataformat-xml.html

docs/oss/security/FasterXML/jackson-dataformat-xml.md
artem-smotrakov: Yes, now it looks correct

https://www.cvedetails.com/vulnerability-list/vendor_id-15866/product_id-61565/Fasterxml-Jackson-dataformat-xml.html

docs/oss/security/FasterXML/jackson-dataformats-text.md
artem-smotrakov: I think so. I could not find CVEs for jackson-dataformats-text.

artem-smotrakov:
> Just an idea just based on my understanding, is it possible to borrow that logic of how Vulas colleagues identify which version fixes it or if there is a fixed version already available. So that out of the list of unpatched vulnerabilities, if for any vulnerability, the fixed version of the library is available, it's a false positive and we can remove it from the list.

We can look into it. Unfortunately, there is no publicly available database which I am aware of.

Sachpat:
> > Just an idea just based on my understanding, is it possible to borrow that logic of how Vulas colleagues identify which version fixes it or if there is a fixed version already available. So that out of the list of unpatched vulnerabilities, if for any vulnerability, the fixed version of the library is available, it's a false positive and we can remove it from the list.
> 
> We can look into it. Unfortunately, there is no publicly available database which I am aware of.

No what I meant is, their logic not the database itself. How do they determine if a particular library version is safe from the detected vulnerability? If I understand correctly, they do some code comparison and commit history right? Will it be too complicated and heavy to include such logic here?

artem-smotrakov:
> > > Just an idea just based on my understanding, is it possible to borrow that logic of how Vulas colleagues identify which version fixes it or if there is a fixed version already available. So that out of the list of unpatched vulnerabilities, if for any vulnerability, the fixed version of the library is available, it's a false positive and we can remove it from the list.
> > 
> > 
> > We can look into it. Unfortunately, there is no publicly available database which I am aware of.
> 
> No what I meant is, their logic not the database itself. How do they determine if a particular library version is safe from the detected vulnerability? If I understand correctly, they do some code comparison and commit history right? Will it be too complicated and heavy to include such logic here?

I don't know much about it but I think they have some automation to find patches but they sometime add links to patches manually. Anyway, it's better to look into it. Maybe we can find some ideas to borrow.

docs/oss/security/apache/activemq.md
sourabhsparkala: Right now it gives `30 vulnerabilities`.

artem-smotrakov: Here I see 28

https://www.cvedetails.com/product/19047/Apache-Activemq.html?vendor_id=45

sourabhsparkala: You know this can be confirmed with #189. As the REST API returns the total number of Vulnerabilities associated to a CPE. Maybe we use that to make this process more intelligent.

artem-smotrakov: Sure, we should definitely look into it.

PR #199, opened 2020-05-25
Removed rating versions

### LINKED ISSUE #187: Better versions for ratings
Currently, `Rating.version()` always return the same value even if the rating has been updated. Therefore, it's not possible to distinguish updates in the rating definition. The `version()` method needs to be updated to return incremental versions when the rating gets updated.

artem-smotrakov:
This closes #187


src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/example/SecurityRatingExample.java
```diff
@@ -44,8 +42,8 @@ public void makeImmutable() {
    * Initializes a security rating with SecurityScoreExample.
    */
   @JsonCreator
```
sourabhsparkala, line 44: Is this still required, if there is no `@JsonProperty`?

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
```diff
@@ -23,10 +22,9 @@
    */
   @JsonCreator
   public OssSecurityRating(
-      @JsonProperty("score") OssSecurityScore score,
-      @JsonProperty("version") Version version) {
+      @JsonProperty("score") OssSecurityScore score) {
 
-    super("Security rating for open-source projects", score, version);
+    super("Security rating for open-source projects", score);
```
sourabhsparkala, line 27: Is this score being null checked?

artem-smotrakov: Yeah. Please see `AbstractRating`.

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/example/SecurityRatingExample.java
artem-smotrakov: Right. I've removed it.

PR #201, opened 2020-05-26
Relative thresholds for labels in the OSS security rating

### LINKED ISSUE #188: Relative thresholds for labels

artem-smotrakov:

- The `OssSecurityRating` can now be configured with thresholds for labels.
- The thresholds are calculated by the `SecurityRatingAnalysis` notepad and stored to a JSON file.
- The `RatingRepository` reads the thresholds from the file and uses them in the open-source security rating.

This closes #188


mibo:
Looks good. 
But there is no way to beside the JSON files to configure the thresholds when the rating is get from the `RatingRepository.rating(..)`, or did I just not found it?

artem-smotrakov:
> But there is no way to beside the JSON files to configure the thresholds when the rating is get from the `RatingRepository.rating(..)`, or did I just not found it?

`RatingRepository` is supposed to return immutable ratings. Therefore, `OssSecurityRating` is immutable. If you'd like to override the thresholds, you can override the `label()` method in a wrapper for the `OssSecurityRating` or `OssSecurityScore` instances that a returned by `RatingRepository`. I can make the `RatingRepository.register()` method public so that the wrapper (new rating) can be added to the repository.

mibo:
> `RatingRepository` is supposed to return immutable ratings. Therefore, `OssSecurityRating` is immutable. 

For me it was that the rating / rating value is immutable but the label (as a simplified view to the rating value) is more flexible, so that it can be easier adjusted based on latest experiences/learnings.
But you are right, as the label is part of the rating and the rating is supposed to be immutable current implementation is fine.

> If you'd like to override the thresholds, you can override the `label()` method in a wrapper for the `OssSecurityRating` or `OssSecurityScore` instances that a returned by `RatingRepository`. I can make the `RatingRepository.register()` method public so that the wrapper (new rating) can be added to the repository.

Sounds a bit cumbersome 🙃
Guess making the register method public is not required as someone can just use the wrapped instance (at least from my current understanding).

Btw. is it intended that the `store(..)` method in the `RatingRepository` is public?

artem-smotrakov:
> > `RatingRepository` is supposed to return immutable ratings. Therefore, `OssSecurityRating` is immutable.
> 
> For me it was that the rating / rating value is immutable but the label (as a simplified view to the rating value) is more flexible, so that it can be easier adjusted based on latest experiences/learnings.
> But you are right, as the label is part of the rating and the rating is supposed to be immutable current implementation is fine.
> 
> > If you'd like to override the thresholds, you can override the `label()` method in a wrapper for the `OssSecurityRating` or `OssSecurityScore` instances that a returned by `RatingRepository`. I can make the `RatingRepository.register()` method public so that the wrapper (new rating) can be added to the repository.
> 
> Sounds a bit cumbersome 

Mapping a score value to a label is a part of a rating. Since a rating is immutable, then this part is also immutable. If you'd like to override it, it means that you create a new rating (even if it is based on another one). The framework allows you to do that.

> Guess making the register method public is not required as someone can just use the wrapped instance (at least from my current understanding).
> 
> Btw. is it intended that the `store(..)` method in the `RatingRepository` is public?

I suspect it should not be public.

PR #202, opened 2020-05-27
Better vulnerability scores

### LINKED ISSUE #191: UnpatchedVulnerabilitiesScore and VulnerabilityLifetimeScore are not reliable
Vulnerability matching doesn't work well, there are a lot of false-positives. For example, see a number of issues for OpenSSL:

https://github.com/artem-smotrakov/fosstars-rating-core/blob/analyse-ratings/src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb

Most of vulnerabilities are considered unpatched. Those are false-positives.

All vulnerabilities don't have dates when they were introduced. Many vulnerabilities don't have dates when they fixed. The existing dates are estimated by the dates when the issues were published. Then, `VulnerabilityLifetimeScore` tries to estimate the dates. As a result, the score becomes very unreliable

What to do:
 - Let's see if #180 improves it. If it doesn't, we may need to disable `VulnerabilityLifetimeScore` that use info about vulnerabilities.
- Try to re-work UnpatchedVulnerabilitiesScore. If it doesn't work out, use only the list of unpatched issues.

artem-smotrakov:
Here is a list of updates:

- Disabled analyzing NVD in `UnpatchedVulnerabilities` data provider.
- Updated `VulnerabilityLifetimeScore` not to estimate dates.
- Removed `VulnerabilityLifetimeScore` from `OssSecurityScore`.
- Disabled `ProjectStarted` provider in the demo tool.
- Updated the existing tests and test vectors.
- Updated docs.

This closes #191

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
@@ -1,6 +1,5 @@
 package com.sap.sgs.phosphor.fosstars.model.score.oss;
 
-import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.PROJECT_START_DATE;
 import static com.sap.sgs.phosphor.fosstars.model.feature.oss.OssFeatures.VULNERABILITIES;
 import static com.sap.sgs.phosphor.fosstars.model.other.Utils.findValue;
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/VulnerabilityLifetimeScore.java
```diff
@@ -75,28 +62,33 @@ static double factorOf(Duration lifetime) {
...
     double penalty = 0.0;
+    boolean found = false;
     for (Vulnerability entry : vulnerabilities.get().entries()) {
-      Instant introduced = entry.introduced().orElse(estimatedIntroducedDate).toInstant();
-      Instant fixed = entry.fixed().orElse(now).toInstant();
+      if (!entry.introduced().isPresent() || !entry.fixed().isPresent()) {
+        continue;
+      }
+
+      found = true;
```
sourabhsparkala, line 88: I think this line should come after #L104. What do you think? 
- The reason being. Consider `if()` in #L93 is `true`. Then the `flag` is already `true` for the next loop.

artem-smotrakov: It makes sense. The flag should be set when we're sure that the dates are correct.

artem-smotrakov:
> * Also could you please check again if all the traces of the removed features are completely removed. I did not find anything as such. But just to be sure.

The juputer notebooks still mention the disabled feature and score. The notebooks are broken anyway, so I didn't want to fix them now. That would make the changeset much larger. We'll fix them separately.

artem-smotrakov:

PR #207, opened 2020-05-29
Fixes ConcurrentModificationException issue in GitHubDataFetcher

### LINKED ISSUE #206: `cleanUp` in `GitHubDataFetcher` leads to exception
During iteration over the `localRepositoriesInfo` the deleted items are removed from the map [see here](https://github.com/SAP/fosstars-rating-core/blob/0fb216dca734b23d34a18eb7e20e13a237ecb26d/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java#L389).

This leads to a `java.util.ConcurrentModificationException`.

Suggestion: in the loop collect deleted items and remove them after the iteration is finished.

mibo:
This closes #206

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -14,9 +14,11 @@ import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import org.apache.commons.collections4.map.LRUMap;
 import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
@@ -375,6 +377,7 @@ public class GitHubDataFetcher {
       total = total.add(info.repositorySize());
     }
 
+    Set<URL> toBeRemoved = new HashSet<>();
     for (Map.Entry<URL, LocalRepositoryInfo> entry : localRepositoriesInfo.entrySet()) {
       URL url = entry.getKey();
       LocalRepositoryInfo info = entry.getValue();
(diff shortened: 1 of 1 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -386,9 +389,10 @@ public void cleanup(CleanupStrategy strategy) {
           LOGGER.error(
               () -> String.format("Could not delete a local repository: %s", info.path()), e);
         }
-        localRepositoriesInfo.remove(url);
+        toBeRemoved.remove(url);
```
artem-smotrakov, line 392: Did you mean to add a URL to the set?

```suggestion
        toBeRemoved.add(url);
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -386,9 +389,10 @@ public void cleanup(CleanupStrategy strategy) {
...
       }
     }
+    toBeRemoved.forEach(localRepositoriesInfo::remove);
```
artem-smotrakov, line 395: `Set.removeAll()` can also be used but not a big deal.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: 🙈 Sure, this would make a lot more sense 🙂

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: Afaik `Set.removeAll()` can not be used at we want to remove the `URL`s from the map (which does not over a `removeAll()` method).

mibo:
Thanks for quick review and merge.

PR #209, opened 2020-05-29
Store only weights of the open-source security score

### LINKED ISSUE #208: Store only weights of the open-source security score
`RatingRepository` loads the open-source security score from `OssSecurityScore_1_0.json`. The file contains a serialized score but the only interesting part is the weights. Instead of storing everything, it's better to store just weights.

artem-smotrakov:
Updated `RatingRepository` to load only weights for the open-source security score.

This closes #208


artem-smotrakov:

PR #212, opened 2020-06-02
Simplify test vectors for scores

### LINKED ISSUE #211: Simplify test vectors for scores
Tests vectors for scores are serialized to YAML and stored in files. The test vectors are serialized "as-is", therefore the files contains a lot of redundant data. That makes it harder to edit the test vectors. Let's simplify them.

artem-smotrakov:

- Added a new `AbstractTestVector` class.
- Added a new `ScoreTestVector` class.
- Added a new `TestVector.valuesFor()` method.
- Updated `ScoreVerified` to use the new `valuesFor()` method.
- Simplified the test vectors for the `OssSecurityScore`.
- Added `ScoreTestVectorTest`.

This closes #211

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
@@ -1,0 +1,106 @@
+package com.sap.sgs.phosphor.fosstars.model.qa;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.sap.sgs.phosphor.fosstars.model.Interval;
+import com.sap.sgs.phosphor.fosstars.model.Label;
+import java.util.Objects;
+
+/**
+ * A base class for test vectors.
+ */
+public abstract class AbstractTestVector implements TestVector {
+
+  /**
+   * An interval for an expected score.
+   */
+  final Interval expectedScore;
+
+  /**
+   * If it's set to true, then a not-applicable score value is expected.
+   */
+  final boolean expectedNotApplicableScore;
+
+  /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
```diff
@@ -0,0 +1,106 @@
...
+   * @param alias A alias of the test vector.
+   * @param expectedNotApplicableScore
+   *        If it's set to true, then a not-applicable score value is expected.
+   */
+  AbstractTestVector(Interval expectedScore, Label expectedLabel, String alias,
+      boolean expectedNotApplicableScore) {
+
+    Objects.requireNonNull(alias, "Hey! alias can't be null!");
+
+    if (expectedScore == null && !expectedNotApplicableScore) {
```
sourabhsparkala, line 47: expectedNotApplicableScore could someone set this as `null`?

artem-smotrakov: Yes, `expectedScore` and `expectedLabel` are allowed to be null. `expectedLabel` may be null when a test vector is for a score (score don't produce labels). `expectedScore` may be null if an N/A value is expected (`expectedNotApplicableScore` should be true).

`expectedNotApplicableScore` can't be null since it's `boolean`.

artem-smotrakov: As a little improvement, we can use `Optional` to avoid null-checks.

artem-smotrakov:

PR #218, opened 2020-06-04
Fix #217: Use synchronized map

### LINKED ISSUE #217: Handling of `localRepositoriesInfo` in `GitHubDataFetcher` is not thread safe.
Handling of `localRepositoriesInfo` in `GitHubDataFetcher` is not thread safe.
If using only one instance of `GitHubDataFetcher` the `cleanup` method throws `ConcurrentModificationException` if a) it is called parallel or b) if during cleanup a new element is added.

Tested with use of a synchronized map and a synchronized block for the clean up.
See PR #218 

In addition I would like to have the option to replace the used `GitHub` instance.
Would it be possible to make it not final and add a `setGitHub` method?
If yes I can create an issue / PR for this 🙂

mibo:
Synchronize access to `localRepositoriesInfo` in `GitHubDataFetcher`.

This closes #217

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java
@@ -2,7 +2,9 @@ package com.sap.sgs.phosphor.fosstars.data.github;
 
 import com.sap.sgs.phosphor.fosstars.data.Cache;
 import com.sap.sgs.phosphor.fosstars.tool.github.GitHubProject;
+import java.util.Collections;
 import java.util.Date;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import org.apache.commons.collections4.map.LRUMap;
@@ -30,13 +32,13 @@ public class GitHubDataCache<T> implements Cache<GitHubProject, T> {
   /**
    * A map of cache entries.
    */
-  private final LRUMap<GitHubProject, GitHubData<T>> entries;
+  private final Map<GitHubProject, GitHubData<T>> entries;
 
   /**
    * The default constructor.
    */
   public GitHubDataCache() {
-    this.entries = new LRUMap<>(CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE);
+    this.entries = Collections.synchronizedMap(new LRUMap<>(CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE));
   }
 
   @Override
@@ -71,7 +73,7 @@ public class GitHubDataCache<T> implements Cache<GitHubProject, T> {
    * Returns the maximum size of the cache.
    */
   public int maxSize() {
-    return entries.maxSize();
+    return CACHE_CAPACITY;
   }
 
   @Override
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -79,9 +79,14 @@ public class GitHubDataFetcher {
   private static final boolean SCAN_UNTIL_REMOVABLE = true;
 
   /**
+   * Lock to synchronize access to localRepositoriesInfo.
+   */
+  private static final Object LOCAL_REPOSITORIES_INFO_LOCK = new Object();
+
+  /**
    * An interface to the GitHub API.
    */
-  private final GitHub github;
+  private GitHub github;
 
   /**
    * A limited capacity cache to store the repository of a {@link GitHubProject}.
@@ -122,8 +127,16 @@ public class GitHubDataFetcher {
   /**
    * Returns the interface to the GitHub API.
    */
-  public GitHub github() {
+  public synchronized GitHub github() {
     return github;
+  }
+
+  /**
+   * Set the interface to the GitHub API.
+   */
+  public synchronized void set(GitHub github) {
+    Objects.requireNonNull(github, "Hey! An interface to GitHub can not be null!");
+    this.github = github;
   }
 
   /**
@@ -210,50 +223,52 @@ public class GitHubDataFetcher {
   private LocalRepository loadLocalRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
 
-    LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
-    if (info == null) {
-      Path repositoryPath = base
-          .resolve(project.organization().name())
-          .resolve(project.name());
+    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+      LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
+      if (info == null) {
+        Path repositoryPath = base
+            .resolve(project.organization().name())
+            .resolve(project.name());
 
-      info = new LocalRepositoryInfo(
-          repositoryPath, Date.from(Instant.now()), project.url());
+        info = new LocalRepositoryInfo(
+            repositoryPath, Date.from(Instant.now()), project.url());
+      }
+
+      if (Files.isRegularFile(info.path())) {
+        LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
+        Files.delete(info.path());
+      }
+
+      Optional<Repository> repository = openRepository(info.path());
+      if (!repository.isPresent()) {
+        Files.deleteIfExists(info.path());
+        clone(project, info.path());
+        repository = openRepository(info.path());
+      }
+
+      if (!repository.isPresent()) {
+        throw new IOException("Could not fetch project's repository!");
+      }
+
+      LocalRepository localRepository = new LocalRepository(info, repository.get());
+
+      if (shouldUpdate(localRepository)) {
+        LOGGER.info("Pulling updates from {} ...", project.url());
+        localRepository.pull();
+      }
+
+      info.updated(Date.from(Instant.now()));
+      localRepositoriesInfo.put(project.url(), info);
+      storeLocalRepositoriesInfo();
+
+      return localRepository;
     }
-
-    if (Files.isRegularFile(info.path())) {
-      LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
-      Files.delete(info.path());
-    }
-
-    Optional<Repository> repository = openRepository(info.path());
-    if (!repository.isPresent()) {
-      Files.deleteIfExists(info.path());
-      clone(project, info.path());
-      repository = openRepository(info.path());
-    }
-
-    if (!repository.isPresent()) {
-      throw new IOException("Could not fetch project's repository!");
-    }
-
-    LocalRepository localRepository = new LocalRepository(info, repository.get());
-
-    if (shouldUpdate(localRepository)) {
-      LOGGER.info("Pulling updates from {} ...", project.url());
-      localRepository.pull();
-    }
-
-    info.updated(Date.from(Instant.now()));
-    localRepositoriesInfo.put(project.url(), info);
-    storeLocalRepositoriesInfo();
-
-    return localRepository;
   }
 
   /**
    * Sets how often new updates should be pulled to a local repository by default.
    */
-  public void pullAfter(Duration duration) {
+  public synchronized void pullAfter(Duration duration) {
     Objects.requireNonNull(duration, "Oh no! Duration is null!");
     pullInterval = duration;
   }
@@ -261,7 +276,7 @@ public class GitHubDataFetcher {
   /**
    * Returns true if a repository should be updated, false otherwise.
    */
-  protected boolean shouldUpdate(LocalRepository repository) {
+  protected synchronized boolean shouldUpdate(LocalRepository repository) {
     Objects.requireNonNull(repository, "Oh no! Repository is null!");
     Instant nextUpdate = repository.info().updated().toInstant().plus(pullInterval);
     return nextUpdate.isBefore(Instant.now());
@@ -334,17 +349,19 @@ public class GitHubDataFetcher {
    * @throws IOException If something went wrong.
    */
   protected Map<URL, LocalRepositoryInfo> loadLocalRepositoriesInfo() throws IOException {
-    Path path = base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE);
+    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+      Path path = base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE);
 
-    if (!Files.exists(path)) {
-      return new HashMap<>();
+      if (!Files.exists(path)) {
+        return new HashMap<>();
+      }
+
+      if (!Files.isRegularFile(path)) {
+        throw new IOException(String.format("Hey! %s is not a file!", path));
+      }
+
+      return MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF);
     }
-
-    if (!Files.isRegularFile(path)) {
-      throw new IOException(String.format("Hey! %s is not a file!", path));
-    }
-
-    return MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF);
   }
 
   /**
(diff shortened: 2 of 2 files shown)


artem-smotrakov:
@mibo You can go ahead and add a `set(GitHub github)` method that updates a `GitHub` instance. Please make it thread-safe. Did you check if the rest of the `GitHubDataFetcher` is thread-safe? I think we need to make it completely thread-safe.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -148,7 +149,7 @@ public GitHubDataFetcher(GitHub github, Path base) throws IOException {
     this.base = base;
     this.github = github;
 
-    this.localRepositoriesInfo = loadLocalRepositoriesInfo();
+    this.localRepositoriesInfo = Collections.synchronizedMap(loadLocalRepositoriesInfo());
```
artem-smotrakov, line 152: Since you you `localRepositoriesInfo` is a synchronized block, is it necessary to use a synchronized map?

mibo: I would say yes, otherwise it may lead to that in the `localRepositoryFor` method new items are added during the `cleanup` run.

mibo:
> @mibo You can go ahead and add a `set(GitHub github)` method that updates a `GitHub` instance. Please make it thread-safe. 

Should I create a separate issue / PR for this?

> Did you check if the rest of the `GitHubDataFetcher` is thread-safe? I think we need to make it completely thread-safe.

Had a look and the rest looks fine for me. 
In addition I checked the `GitHubDataCache` and looks also good.

artem-smotrakov:
> > @mibo You can go ahead and add a `set(GitHub github)` method that updates a `GitHub` instance. Please make it thread-safe.
> 
> Should I create a separate issue / PR for this?

Updating this pull request is fine.

> 
> > Did you check if the rest of the `GitHubDataFetcher` is thread-safe? I think we need to make it completely thread-safe.
> 
> Had a look and the rest looks fine for me.
> In addition I checked the `GitHubDataCache` and looks also good.

Okay, let me have another look.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: Then, should those methods also access the map in a synchronized block? I think that using both synchronized blocks and a synchronized map is redundand.

artem-smotrakov:
> > Did you check if the rest of the `GitHubDataFetcher` is thread-safe? I think we need to make it completely thread-safe.
> 
> Had a look and the rest looks fine for me.
> In addition I checked the `GitHubDataCache` and looks also good.

I think the rest of fields in the class need to be synchronized as well to make the class thread-safe.

mibo:
> > Had a look and the rest looks fine for me.
> > In addition I checked the `GitHubDataCache` and looks also good.
> 
> I think the rest of fields in the class need to be synchronized as well to make the class thread-safe.

Hm, I'm not the expert for concurrency but in `GitHubDataCache` the `entries` (`LRUMap`) is used in a way that I expect in worst case that the fetcher use a cache entry which in parallel gets removed.
Because the `GitHubDataCache.get(..)` anyways return `Optional`.

To be on the very safe side we can wrap the `entries` map:
`this.entries = Collections.synchronizedMap(new LRUMap<>(CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE));`

WDYT?

artem-smotrakov:
The `repositoryFor()` method reads and writes data to `repositoryCache`. I think it may cause a race. Therefore, I think that `repositoryCache` needs to be accessed in a synchronized block or `GitHubDataCache` needs to be made thread-safe.

The same is for `localRepositoriesInfo` and `localRepositories`.

Furthermore, the `loadLocalRepositoryFor()` method works with the local repositories. That also needs to be synchronized between multiple instances of the class since all of them use the same directory.

mibo:
> The `repositoryFor()` method reads and writes data to `repositoryCache`. I think it may cause a race. Therefore, I think that `repositoryCache` needs to be accessed in a synchronized block or `GitHubDataCache` needs to be made thread-safe.

As mentioned guess the worst what can happen is that in the `repositoryFor` method a `GHRepository` is used which just got removed from the cache.
But to be on the safe side the `LRUMap` in `GitHubDataCache` can be wrapped.

> The same is for `localRepositoriesInfo` and `localRepositories`.
> 
> Furthermore, the `loadLocalRepositoryFor()` method works with the local repositories. That also needs to be synchronized between multiple instances of the class since all of them use the same directory.

Yes, I missed this as I confused myself with `loadLocalRepositoriesInfo` and `loadLocalRepositoryFor`.
Agree that a `synchronized` block should be added to `loadLocalRepositoryFor` and the `storeLocalRepositoriesInfo`.
For the `loadLocalRepositoriesInfo` guess it is not required as it is only called in the constructor.
But as it can be called by extending classes guess it is better to add a `synchronized` block.

Will update the PR accordingly.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: Yes, with adding the additional `synchronized` blocks the wrapping is not mandatory.
But removing it also does not really change anything regarding performance (afaik).
And keeping it makes it at least clear / safe in case of modifications in the future.

artem-smotrakov: > Yes, with adding the additional `synchronized` blocks the wrapping is not mandatory.

Then, let's please avoid duplicating synchronization.

> But removing it also does not really change anything regarding performance (afaik).

Most likely we won't notice that, but it does.

> And keeping it makes it at least clear / safe in case of modifications in the future.

I am sorry but I don't see how it make it clearer and safer. Let's please avoid duplicating synchronization.

mibo: > > But removing it also does not really change anything regarding performance (afaik).
> 
> Most likely we won't notice that, but it does.

Did not know this. Always thought performance is equal if the `synchronized` block is `synchronized` on the same object which is wrapped via the `Collections.synchronized...`.
And in our case the `localRepositoriesInfo`.
But thanks for this info 🙂

> > And keeping it makes it at least clear / safe in case of modifications in the future.
> 
> I am sorry but I don't see how it make it clearer and safer. Let's please avoid duplicating synchronization.

Done.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java
```diff
@@ -71,7 +73,7 @@ public int size() {
    * Returns the maximum size of the cache.
    */
   public int maxSize() {
-    return entries.maxSize();
+    return CACHE_CAPACITY;
```
artem-smotrakov, line 76: Is this update required?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -126,6 +127,14 @@ public GitHub github() {
     return github;
   }
 
+  /**
+   * Set the interface to the GitHub API.
+   */
+  public void github(GitHub github) {
```
artem-smotrakov, line 133: Could you please rename it to `set()`? I think that the `github` field needs also be synchronized.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -210,44 +219,46 @@ public LocalRepository localRepositoryFor(GitHubProject project) throws IOExcept
   private LocalRepository loadLocalRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
 
-    LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
-    if (info == null) {
-      Path repositoryPath = base
-          .resolve(project.organization().name())
-          .resolve(project.name());
+    synchronized (localRepositoriesInfo) {
```
artem-smotrakov, line 222: I am afraid that using a monitor of the `localRepositoriesInfo` field is not enough. The `loadLocalRepositoryFor()` method accesses cloned repositories and keep info about them in the `localRepositoriesInfo` map. When an instance of `GitHubDataFetcher` is used in multiple threads, then a lock on `localRepositoriesInfo` is going to protect the map from a race. However, when we have multiple instances of `GitHubDataFetcher`, then calling the`loadLocalRepositoryFor()` method on them is going to result concurrent access to the filesystem. I think the method should use a monitor of a static constant that is shared between all instances of the class. Or, the method can use a lock-file in the filesystem.

mibo: Perhaps a bit late to ask this. 
But do you think about making it thread-safe for use as singleton only would be better?
In combination with adding according static method to get (create) the singleton instance?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: I wanted to name it `setGithub` but after I have seen the `github()` method without `get...` I thought without `set` is preferred 🙂
Regarding synchronized for the github field I guess I do not understand it.
Would it make a real difference if the get/set for `github` have a synchronized block?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java
mibo: With the wrapping of the LRUMap I had to change the `entrie` field to `Map`.
As result I need to cast the `entries` to access the max size method.
Or I thought the max size is always the same as the static final value of `CACHE_CAPACITY`.

But I'm open for other options.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: > I wanted to name it `setGithub` but after I have seen the `github()` method without `get...` I thought without `set` is preferred 

I meant `public void set(GitHub github) {`.

> Regarding synchronized for the github field I guess I do not understand it.
> Would it make a real difference if the get/set for `github` have a synchronized block?

Since the `github` field is not final anymore, multiple threads can read and write to the `github` field. Therefore, all access to this field needs to be synchronized.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: I am not sure what you mean and why it is late. Do you suggest to make it a singleton?

I suggest to use the following:

```
private static final Object localRepositoryLock = new Object();
...
synchronized (localRepositoryLock) {
  ... work with local repositories
}
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java
artem-smotrakov: Okay.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: No, I do not see a problem. 
Just thought make `GitHubDataFetcher` is less effort 😄 
Will update it accordingly.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: Still not sure if this would make a bit difference.
But I will update it as proposed and for a detailed explanation we use another channel 🙂

artem-smotrakov: Most likely you won't see any difference. If `github` is not synchronized, and you use multiple instances of `GitHub`, then you can't tell which one is used for sure once you update the field. I would like to make the class completely thread-safe.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: Sure.

> Just thought make GitHubDataFetcher is less effort

Sorry, didn't get it. To make it what?

mibo: Sorry, missed _singleton_

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: Still not sure about this. 
But I trust you as you have more knowledge about concurrency than I 🙂

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: No, I am not sure if making it a singleton would be less effort. This change would result to updating all data provider and their tests.

mibo: Yes, you are right. Have not thought about this.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -210,58 +223,60 @@ public LocalRepository localRepositoryFor(GitHubProject project) throws IOExcept
   private LocalRepository loadLocalRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
 
-    LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
-    if (info == null) {
-      Path repositoryPath = base
-          .resolve(project.organization().name())
-          .resolve(project.name());
+    synchronized (LOCAL_REPOSITORIES_LOCK) {
+      LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
```
artem-smotrakov, line 227: Strictly speaking, `localRepositoriesInfo` should be static. But let's leave it as-is for now. We'll consider this update later.

artem-smotrakov:
> Could you please synchronize `pullInterval` and `github` as well? I think if we should make the class completely thread-safe.

@mibo

artem-smotrakov:
@mibo Do you have a test for the updates?

mibo:
> @mibo Do you have a test for the updates?

Not yet, thought we should first finish the changes in the class(es) and than I would add the tests.

mibo:
> > Could you please synchronize `pullInterval` and `github` as well? I think if we should make the class completely thread-safe.
> 
> @mibo

Is it not enough to add the `synchronized` to the get/set methods?
Guess I have to read some more about concurrency in java to fully understand it.

artem-smotrakov:
> > > Could you please synchronize `pullInterval` and `github` as well? I think if we should make the class completely thread-safe.
> > 
> > 
> > @mibo
> 
> Is it not enough to add the `synchronized` to the get/set methods?
> Guess I have to read some more about concurrency in java to fully understand it.

My bad, I noticed that you added `synchronized` to the methods after I posted my comment.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -81,13 +81,14 @@
...
    */
-  private final GitHub github;
+  private GitHub github;
 
   /**
    * A limited capacity cache to store the repository of a {@link GitHubProject}.
    */
   private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
 
+
```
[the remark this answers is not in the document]
mibo: Done

artem-smotrakov:
> > @mibo Do you have a test for the updates?
> 
> Not yet, thought we should first finish the changes in the class(es) and than I would add the tests.

Okay, up to you. I usually add tests right away to make sure that the new stuff work fine.

artem-smotrakov:
@mibo Thanks for the improving `GitHubDataFetcher`. The updates are in master now. I just updated the commit message to refer to the issue.

Issue #192, opened 2020-05-14
Introduce StaticAnalysisScore


artem-smotrakov:
Both `LgtmScore` and `FindSecBugsScore` assess how a project uses static analysis. It's not necessary for a project to use both (or even more tools). The scores may be combined to a `StaticAnalysisScore`. Let's add `StaticAnalysisScore`.

PR #222, opened 2020-06-10
Added StaticAnalysisScore

artem-smotrakov:
This is a part of #192
 
- Added a new `StaticAnalysisScore` that is based on `LgtmScore` and `FindSecBugsScore`.
- Added a couple of helper methods in `AbstractScore`.
- Updated `LgtmScore` and `FindSecBugsScore` to return unknown and N/A values if necessary.
- Added `StaticAnalysisScore.Verification`.
- Added test vectors for `StaticAnalysisScore`.
- Added `StaticAnalysisScoreVerification` test and updated other tests.

Next, I'm planning to add the new score to the open-source security score.

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
@@ -26,6 +26,9 @@ import org.apache.logging.log4j.Logger;
  */
 public abstract class AbstractScore implements Score {
 
+  /**
+   * No description.
+   */
   static final String EMPTY_DESCRIPTION = "";
 
   /**
@@ -50,7 +53,7 @@ public abstract class AbstractScore implements Score {
    *
    * @param name A name of the score.
    */
-  AbstractScore(String name) {
+  public AbstractScore(String name) {
     this(name, EMPTY_DESCRIPTION);
   }
 
@@ -60,7 +63,7 @@ public abstract class AbstractScore implements Score {
    * @param name A name of the score.
    * @param description A description of the score (may be empty).
    */
-  AbstractScore(String name, String description) {
+  public AbstractScore(String name, String description) {
     Objects.requireNonNull(name, "Hey! Score name can't be null!");
     Objects.requireNonNull(description, "Hey! Score description can't be null!");
     if (name.isEmpty()) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
```diff
@@ -240,4 +252,50 @@ static ScoreValue calculateIfNecessary(Score score, ValueHashSet values) {
...
   }
+
+  /**
+   * Checks if all values are unknown.
+   *
+   * @param values The values to be checked.
+   * @return True if all values are unknown, false otherwise.
+   * @throws IllegalArgumentException If values are empty.
+   */
+  protected static boolean allUnknown(Value... values) {
```
Sachpat, line 263: I would name this method as areAllUnknown()

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
```diff
@@ -240,4 +252,50 @@ static ScoreValue calculateIfNecessary(Score score, ValueHashSet values) {
...
+  }
+
+  /**
+   * Checks if all values are N/A.
+   *
+   * @param values The values to be checked.
+   * @return True if all values are N/A, false otherwise.
+   * @throws IllegalArgumentException If values are empty.
+   */
+  protected static boolean allNotApplicable(Value... values) {
```
Sachpat, line 286: The same areAllNotApplicable()

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/AbstractScore.java
artem-smotrakov: I would do that as well if it was not a static method. Then, client code would look like this:

```java
if (values.areAllUnknown()) {
```

Here we have a static method, and client code looks like the following:

```java
if (allUnknown(values)) {
```

It looks fine to me. Please let me know if you see any issue here. Otherwise, let's please keep it as-is.

Sachpat: OK )

artem-smotrakov:

PR #226, opened 2020-06-15
New feature for bug bounty programs

### LINKED ISSUE #102: Introduce a feature for bug bounties for open-source projects
There are a number bug bounties for open-source projects, for example see [Internet Bug Bounties on HackeOne](https://hackerone.com/directory/programs?ibb=true&order_direction=DESC&order_field=started_accepting_at). If a project has such a bug bounty, then it may bring more researchers to look for security issues and report them responsibly. In the end, it improves security.

Let's introduce a boolean feature which shows if a project has a bug bounty. The feature can contribute to the security testing score or directly to the security score (in both cases, it needs an intermediate sub-score since both score are based on sub-scores).

Here is a list of well-known platforms which host bug bounties:

- https://hackerone.com
- https://bugcrowd.com/programs

- [x] Add the new feature.
- [x] Implement a data provider which fills out the new feature. It would be best, if the provider can fetch the info from the bug bounty platforms to fill out the feature. Otherwise, the provider can read a file which maps projects to bug bounties. The file will need to be updated manually which is not very convenient.
- [ ] Include the new feature to the open-source security rating.
- [ ] Update test and test vectors.
- [ ] Update the demo tool.

### LINKED ISSUE #227: Remove duplicate load() methods from JSON storages
The classes in the `com.sap.sgs.phosphor.fosstars.data.json` package use the same mechanism to load data. However, they have duplicate code. Let's fix it.

artem-smotrakov:

- Added a new `OssFeatures.HAS_BUG_BOUNTY_PROGRAM` feature.
- Updated `ProjectSecurityAwarenessScore` to take into account the new feature.
- Added a new data provider `HasBugBountyProgram`. The provider uses a static list of known bug bounties.
- Updated `PrettyPrinter` to print out the new feature.
- Added tests and test vectors.
- Removed duplicate `load()` methods from JSON storages. This closes #227.

This closes #102

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/json/BugBountyProgramStorage.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/json/BugBountyProgramStorage.java
@@ -2,9 +2,7 @@ package com.sap.sgs.phosphor.fosstars.data.json;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.sap.sgs.phosphor.fosstars.tool.github.GitHubProject;
-import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.URL;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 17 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/json/BugBountyProgramStorage.java
```diff
@@ -0,0 +1,121 @@
...
+    return load(RESOURCE_PATH);
+  }
+
+  /**
+   * Loads info about bug bounty programs from a specified file.
+   *
+   * @return An instance of {@link BugBountyProgramStorage}.
+   * @throws IOException If something went wrong.
+   */
+  public static BugBountyProgramStorage load(String path) throws IOException {
```
sourabhsparkala, line 84: This load of file function is used a lot of time. Multiple classes.
Do you think it would be a good idea to have a generic static method for every class?

artem-smotrakov: That's definitely a good idea. I've moved the load method to the abstract parent.

artem-smotrakov:

PR #231, opened 2020-06-18
Update security ratings for well-known projects (0.9.0)

### LINKED ISSUE #225: Make data providers more confident
The `HasSecurityTeam` provider returns an unknown value if a list of known security teams doesn't have the project. It affects the confidence of the rating value. Let's update the provider to set the feature to False if no security team found.

### LINKED ISSUE #228: Update security ratings for well-known projects (0.9.0)
Let's re-calculate security ratings for well-known open-source project with the new version 0.9.0.

### LINKED ISSUE #229: Updated LgtmDataProvider to check N latest commits
Currently the `LgtmDataProvider` checks if LGTM analysed commits in last three month. If a project is not very active, then this behavior may result to a false-negative. Instead, the provider can check latest N commits.

### LINKED ISSUE #230: PrettyPrinter should print sorted values
The report created by `PrettyPrinter` would look a bit better if it sorted values before printing them out (for example, programming languages and package managers). That would also result to less changes in the report for well-known open-source projects when it's updated.

artem-smotrakov:

- Made `HasSecurityTeam` provider more confident. This closes #225.
- Updated `LgtmDataProvider` to check N latest commits. This closes #229.
- Updated `PrettyPrinter` to print sorted values. This closes #230.

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -99,11 +99,6 @@ public class GitHubDataFetcher {
   private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
 
   /**
-   * A limited capacity cache to store all the commits of a {@link GitHubProject}.
-   */
-  private final GitHubDataCache<List<GHCommit>> githubCommitsCache = new GitHubDataCache<>();
-
-  /**
    * A cache of local repositories.
    */
   final LRUMap<GitHubProject, LocalRepository> localRepositories
@@ -175,34 +170,29 @@ public class GitHubDataFetcher {
   }
 
   /**
-   * Returns the cache with GitHub commits.
-   */
-  public GitHubDataCache<List<GHCommit>> githubCommitsCache() {
-    return githubCommitsCache;
-  }
-
-  /**
-   * Returns a list of commits for a GitHub project using GitHub API.
+   * Returns a number of latest commits.
    *
    * @param project The project.
-   * @param duration A time frame.
-   * @return The list of commits
+   * @param n The number of commits.
+   * @return The list of commits.
    * @throws IOException If something went wrong.
    */
-  public List<GHCommit> githubCommitsFor(GitHubProject project, Duration duration)
-      throws IOException {
-
-    Optional<List<GHCommit>> cachedCommits = githubCommitsCache.get(project);
-    if (cachedCommits.isPresent()) {
-      return cachedCommits.get();
+  public List<GHCommit> githubCommitsFor(GitHubProject project, int n) throws IOException {
+    Objects.requireNonNull(project, "Oh no! The project is null!");
+    if (n <= 0) {
+      throw new IllegalArgumentException("Oh no! The number of commit is not positive!");
     }
 
     try {
-      Date date = Date.from(Instant.now().minus(duration));
       List<GHCommit> commits = new ArrayList<>();
-      for (GHCommit commit : repositoryFor(project).queryCommits().since(date).list()) {
+      for (GHCommit commit : repositoryFor(project).listCommits()) {
         commits.add(commit);
+        n--;
+        if (n == 0) {
+          break;
+        }
       }
+
       return commits;
     } catch (HttpException e) {
       LOGGER.error(String.format("Could not fetch commits from %s", project.url()), e);
@@ -224,7 +214,7 @@ public class GitHubDataFetcher {
       return cachedRepository.get();
     }
 
-    GHRepository repository = github.getRepository(project.path());
+    GHRepository repository = github().getRepository(project.path());
     if (repository == null) {
       throw new IOException(String.format("Could not fetch repository %s (null)", project.url()));
     }
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/HasSecurityTeam.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/HasSecurityTeam.java
@@ -39,9 +39,6 @@ public class HasSecurityTeam extends CachedSingleFeatureGitHubDataProvider {
   @Override
   protected Value fetchValueFor(GitHubProject project) throws IOException {
     logger.info("Figuring out if the project has a security team ...");
-    if (securityTeam.existsFor(project.url())) {
-      return HAS_SECURITY_TEAM.value(true);
-    }
-    return HAS_SECURITY_TEAM.unknown();
+    return HAS_SECURITY_TEAM.value(securityTeam.existsFor(project.url()));
   }
 }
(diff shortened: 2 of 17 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -175,34 +170,29 @@ public GitHubDataFetcher(GitHub github, Path base) throws IOException {
...
+    Objects.requireNonNull(project, "Oh no! The project is null!");
+    if (n <= 0) {
+      throw new IllegalArgumentException("Oh no! The number of commit is not positive!");
     }
 
     try {
-      Date date = Date.from(Instant.now().minus(duration));
       List<GHCommit> commits = new ArrayList<>();
-      for (GHCommit commit : repositoryFor(project).queryCommits().since(date).list()) {
+      for (GHCommit commit : repositoryFor(project).listCommits()) {
```
sourabhsparkala, line 188: Are these commits sorted? in reverse order?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -175,34 +170,29 @@ public GitHubDataFetcher(GitHub github, Path base) throws IOException {
...
         commits.add(commit);
+        n--;
+        if (n == 0) {
+          break;
```
sourabhsparkala, line 192: Does it make sense to use `dowhile` loop here? I remember you suggesting something similar to me. To avoid `break` maybe? I do not remember exactly.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/HasSecurityTeam.java
```diff
@@ -39,9 +39,6 @@ protected Feature supportedFeature() {
   @Override
   protected Value fetchValueFor(GitHubProject project) throws IOException {
     logger.info("Figuring out if the project has a security team ...");
-    if (securityTeam.existsFor(project.url())) {
-      return HAS_SECURITY_TEAM.value(true);
-    }
-    return HAS_SECURITY_TEAM.unknown();
```
sourabhsparkala, line 45: I think this should be `unknown` right?

As we check this feature based on a file which we update on a regular basis. If the security team entry is not present for a project in the file. It doesn't mean that we can confidently say that the project doesn't have a security team. It would be `unknown` right?

Please correct me if I am mistaken.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: Yes, the comments are sorted by the commit date. Latest commits go first.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Languages.java
```diff
@@ -120,7 +120,8 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return elements.stream().map(Enum::toString).collect(Collectors.joining(", "));
+    return elements.stream()
```
sourabhsparkala, line 123: There seems to be no difference in the line.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/PackageManagers.java
```diff
@@ -144,7 +144,8 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return packageManagers.stream().map(Enum::toString).collect(Collectors.joining(", "));
+    return packageManagers.stream()
+        .sorted().map(Enum::toString).collect(Collectors.joining(", "));
```
sourabhsparkala, line 148: There seems to be no difference in the line.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: I though about a do-while loop here. If we use a do-while loop, we'll need to check if the next element in the commit list exist, so the `break` statement will be still there. I think it's a bit shorter to use a for-loop.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/PackageManagers.java
artem-smotrakov: I added `sorted()` call.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/format/PrettyPrinter.java
```diff
@@ -46,42 +46,18 @@
...
-    FEATURE_CLASS_TO_NAME.put(
-        UsesFindSecBugs.class,
-        "Uses FindSecBugs");
-    FEATURE_CLASS_TO_NAME.put(
-        UsesNoHttpTool.class,
-        "Uses nohttp");
-    FEATURE_CLASS_TO_NAME.put(
-        DependencyScanScore.class,
-        "Dependency testing");
+    FEATURE_CLASS_TO_NAME.put(OssSecurityScore.class, "Security of project");
```
sourabhsparkala, line 49: Is this format alright? I liked the earlier one. It is just my preference to be readable.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/PackageManagers.java
artem-smotrakov: But looks like this call is not necessary anymore since the class uses `TreeSet`. I'll remove it.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: ok

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/PackageManagers.java
sourabhsparkala: 👍

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/HasSecurityTeam.java
artem-smotrakov: You are absolutely right. However, returning an unknown value makes overall confidence of rating lower. I am not sure if it is worth it. See #225

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Languages.java
artem-smotrakov: I added sorted() call. But looks like this call is not necessary anymore since the class uses TreeSet. I'll remove it.

src/main/java/com/sap/sgs/phosphor/fosstars/tool/format/PrettyPrinter.java
artem-smotrakov: Yes, this just updates formatting. The number of features and score are going to grow. Let's make the code a bit more compact.

artem-smotrakov:
Thanks for the review @sourabhsparkala ! I've addressed your comments and added a couple of tests.

PR #246, opened 2020-06-30
Make GitHubDataFetcher thread safe

### LINKED ISSUE #210: Update of required after cleanup in GitHubDataFetcher
During some test with `GitHubDataFetcher` and the `cleanup()` method it looks like that use of more than one instance of `GitHubDataFetcher` leads to invalid/outdated data in `localRepositoriesInfo`

As if instance A of `GitHubDataFetcher` runs the `cleanup()` and afterwards a new instance B is created it loads the outdated information via `loadLocalRepositoriesInfo()` in the c'tor.

Perhaps after the clean up is finished [see here](https://github.com/SAP/fosstars-rating-core/blob/0d28e118ddbf158a10efe38d4828a629a3371b0c/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java#L396) the `storeLocalRepositoriesInfo()` should be called (and a `synchronized` be added to the `storeLocalRepositoriesInfo()`).

Maybe this is not a harmful bug or not a bug at all, but at least I wanted to ask 🙂

### LINKED ISSUE #221: Change map(s) in GitHubDataFetcher to static
At least the [localRepositoriesInfo ](https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java#L120) map in `GitHubDataFetcher` should be static [see comment here](https://github.com/SAP/fosstars-rating-core/pull/218#discussion_r435899992) which I unfortunately missed in the PR.

Additional I would question if [localRepositories](https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java#L109) should be static as well.

As I missed it in my PR I would volunteer to create a PR (next week).
If someone has time earlier feel free to create a PR earlier.

sourabhsparkala:
- `localRepositoriesInfo` has to be set to `static`.
  - Also, with new `GitHubDataFetcher` instance, this Map is loaded with the current list of local repositories. As expected.
- `storeLocalRepositoriesInfo()` is required at the end of `cleanup()`
- Add multiple test case scenarios
  - Update GitHub instance and confirm if fetcher is using the new GitHub instance
  - Load a repository and clean up the repository
  - Clean up the repository without loading the repository
  - Test multi-thread access to synchronized methods
  - Test multi-thread parallel access to a single synchronized method
  - Check if localRepositoriesInfo.json file is being updated on each load and clean up tasks

This fixes #210 
This fixes #221

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -385,8 +385,8 @@ public class GitHubDataFetcher {
         throw new IOException(String.format("Hey! %s is not a file!", path));
       }
 
-      LOCAL_REPOSITORIES_INFO
-          .putAll(MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
+      LOCAL_REPOSITORIES_INFO.putAll(
+          MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
     }
   }
 
@@ -395,7 +395,7 @@ public class GitHubDataFetcher {
    *
    * @throws IOException If something went wrong.
    */
-  protected static void storeLocalRepositoriesInfo() throws IOException {
+  private static void storeLocalRepositoriesInfo() throws IOException {
     synchronized (LOCAL_REPOSITORIES_INFO) {
       Files.write(
           DEFAULT_PATH_LOCAL_REPOSITORIES_INFO_FILE,
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -37,9 +37,13 @@ import org.kohsuke.github.GitHub;
 import org.kohsuke.github.HttpException;
 
 /**
+ * <p>
  * Helper class for GitHub data providers which pulls the data from a GitHub repository. Also, the
  * class caches the fetched data for a certain amount of time.
- * <p>The class is thread-safe.</p>
+ * </p>
+ * <p>
+ * The class is thread-safe.
+ * </p>
  */
 public class GitHubDataFetcher {
 
@@ -67,17 +71,18 @@ public class GitHubDataFetcher {
   /**
    * The default base directory.
    */
-  private static final String DEFAULT_DIRECTORY = ".fosstars/repositories";
+  static final Path DEFAULT_DIRECTORY = Paths.get(".fosstars/repositories");
 
   /**
    * The default file for info about local repositories.
    */
-  private static final String DEFAULT_LOCAL_REPOSITORIES_INFO_FILE = "local_repositories_info.json";
+  private static final Path DEFAULT_PATH_LOCAL_REPOSITORIES_INFO_FILE
+      = DEFAULT_DIRECTORY.resolve("local_repositories_info.json");
 
   /**
    * Maximum size of the cache for local repositories.
    */
-  private static final int LOCAL_REPOSITORIES_CACHE_CAPACITY = 100;
+  static final int LOCAL_REPOSITORIES_CACHE_CAPACITY = 100;
 
   /**
    * This flag doesn't allow exceeding the maximum cache size.
@@ -85,20 +90,25 @@ public class GitHubDataFetcher {
   private static final boolean SCAN_UNTIL_REMOVABLE = true;
 
   /**
-   * Lock to synchronize access to localRepositoriesInfo.
+   * A synchronized cache of local repositories.
    */
-  private static final Object LOCAL_REPOSITORIES_INFO_LOCK = new Object();
+  static final Map<GitHubProject, LocalRepository> LOCAL_REPOSITORIES
+      = Collections.synchronizedMap(
+        new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE));
 
   /**
-   * A cache of local repositories.
+   * Synchronized map containing info about local repositories.
    */
-  static final LRUMap<GitHubProject, LocalRepository> LOCAL_REPOSITORIES
-      = new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE);
+  static final Map<URL, LocalRepositoryInfo> LOCAL_REPOSITORIES_INFO
+      = Collections.synchronizedMap(new HashMap<>());
 
-  /**
-   * Info about local repositories.
-   */
-  static final Map<URL, LocalRepositoryInfo> LOCAL_REPOSITORIES_INFO = new HashMap<>();
+  static {
+    try {
+      loadLocalRepositoriesInfo();
+    } catch (IOException e) {
+      throw new ExceptionInInitializerError(e);
+    }
+  }
 
   /**
    * An interface to the GitHub API.
@@ -109,11 +119,6 @@ public class GitHubDataFetcher {
    * A limited capacity cache to store the repository of a {@link GitHubProject}.
    */
   private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
-
-  /**
-   * A base directory.
-   */
-  private final Path base;
 
   /**
    * Defines how often new updates should be pulled to a local repository.
@@ -127,7 +132,17 @@ public class GitHubDataFetcher {
    * @throws IOException If something went wrong.
    */
   public GitHubDataFetcher(GitHub github) throws IOException {
-    this(github, Paths.get(DEFAULT_DIRECTORY));
+    Objects.requireNonNull(github, "Hey! An interface to GitHub can not be null!");
+
+    if (!Files.exists(DEFAULT_DIRECTORY)) {
+      Files.createDirectories(DEFAULT_DIRECTORY);
+    }
+
+    if (!Files.isDirectory(DEFAULT_DIRECTORY)) {
+      throw new IOException(String.format("Hey! %s is not a directory!", DEFAULT_DIRECTORY));
+    }
+
+    this.github = github;
   }
 
   /**
@@ -143,31 +158,6 @@ public class GitHubDataFetcher {
   public synchronized void set(GitHub github) {
     Objects.requireNonNull(github, "Hey! An interface to GitHub can not be null!");
     this.github = github;
-  }
-
-  /**
-   * Initializes a new data fetcher.
-   *
-   * @param github An interface to the GitHub API.
-   * @param base A base directory.
-   * @throws IOException If something went wrong.
-   */
-  public GitHubDataFetcher(GitHub github, Path base) throws IOException {
-    Objects.requireNonNull(github, "Hey! An interface to GitHub can not be null!");
-    Objects.requireNonNull(base, "Hey! Base directory can not be null!");
-
-    if (!Files.exists(base)) {
-      Files.createDirectories(base);
-    }
-
-    if (!Files.isDirectory(base)) {
-      throw new IOException(String.format("Hey! %s is not a directory!", base));
-    }
-
-    this.base = base;
-    this.github = github;
-
-    loadLocalRepositoriesInfo();
   }
 
   /**
@@ -240,15 +230,14 @@ public class GitHubDataFetcher {
    */
   public LocalRepository localRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
-      LocalRepository repository = LOCAL_REPOSITORIES.get(project);
-      if (repository == null) {
-        repository = loadLocalRepositoryFor(project);
-        LOCAL_REPOSITORIES.put(project, repository);
-      }
+    LocalRepository repository = LOCAL_REPOSITORIES.get(project);
 
-      return repository;
+    if (repository == null) {
+      repository = loadLocalRepositoryFor(project);
+      LOCAL_REPOSITORIES.put(project, repository);
     }
+
+    return repository;
   }
 
   /**
@@ -261,41 +250,44 @@ public class GitHubDataFetcher {
   private LocalRepository loadLocalRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
 
-    LocalRepositoryInfo info = LOCAL_REPOSITORIES_INFO.get(project.url());
-    if (info == null) {
-      Path repositoryPath = base.resolve(project.organization().name()).resolve(project.name());
+    synchronized (LOCAL_REPOSITORIES_INFO) {
+      LocalRepositoryInfo info = LOCAL_REPOSITORIES_INFO.get(project.url());
+      if (info == null) {
+        Path repositoryPath =
+            DEFAULT_DIRECTORY.resolve(project.organization().name()).resolve(project.name());
 
-      info = new LocalRepositoryInfo(repositoryPath, Date.from(Instant.now()), project.url());
+        info = new LocalRepositoryInfo(repositoryPath, Date.from(Instant.now()), project.url());
+      }
+
+      if (Files.isRegularFile(info.path())) {
+        LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
+        Files.delete(info.path());
+      }
+
+      Optional<Repository> repository = openRepository(info.path());
+      if (!repository.isPresent()) {
+        Files.deleteIfExists(info.path());
+        clone(project, info.path());
+        repository = openRepository(info.path());
+      }
+
+      if (!repository.isPresent()) {
+        throw new IOException("Could not fetch project's repository!");
+      }
+
+      LocalRepository localRepository = new LocalRepository(info, repository.get());
+
+      if (shouldUpdate(localRepository)) {
+        LOGGER.info("Pulling updates from {} ...", project.url());
+        localRepository.pull();
+      }
+
+      info.updated(Date.from(Instant.now()));
+      LOCAL_REPOSITORIES_INFO.put(project.url(), info);
+      storeLocalRepositoriesInfo();
+
+      return localRepository;
     }
-
-    if (Files.isRegularFile(info.path())) {
-      LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
-      Files.delete(info.path());
-    }
-
-    Optional<Repository> repository = openRepository(info.path());
-    if (!repository.isPresent()) {
-      Files.deleteIfExists(info.path());
-      clone(project, info.path());
-      repository = openRepository(info.path());
-    }
-
-    if (!repository.isPresent()) {
-      throw new IOException("Could not fetch project's repository!");
-    }
-
-    LocalRepository localRepository = new LocalRepository(info, repository.get());
-
-    if (shouldUpdate(localRepository)) {
-      LOGGER.info("Pulling updates from {} ...", project.url());
-      localRepository.pull();
-    }
-
-    info.updated(Date.from(Instant.now()));
-    LOCAL_REPOSITORIES_INFO.put(project.url(), info);
-    storeLocalRepositoriesInfo();
-
-    return localRepository;
   }
 
   /**
@@ -381,14 +373,15 @@ public class GitHubDataFetcher {
    * @return The info about local repositories.
    * @throws IOException If something went wrong.
    */
-  protected void loadLocalRepositoriesInfo() throws IOException {
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
-      Path path = base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE);
-
+  static void loadLocalRepositoriesInfo() throws IOException {
+    Path path = DEFAULT_PATH_LOCAL_REPOSITORIES_INFO_FILE;
+    synchronized (LOCAL_REPOSITORIES_INFO) {
       if (!Files.exists(path)) {
         LOCAL_REPOSITORIES_INFO.clear();
         return;
-      } else if (!Files.isRegularFile(path)) {
+      }
+
+      if (!Files.isRegularFile(path)) {
         throw new IOException(String.format("Hey! %s is not a file!", path));
       }
 
@@ -402,10 +395,10 @@ public class GitHubDataFetcher {
    *
    * @throws IOException If something went wrong.
    */
-  protected void storeLocalRepositoriesInfo() throws IOException {
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+  protected static void storeLocalRepositoriesInfo() throws IOException {
+    synchronized (LOCAL_REPOSITORIES_INFO) {
       Files.write(
-          base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE),
+          DEFAULT_PATH_LOCAL_REPOSITORIES_INFO_FILE,
           MAPPER.writeValueAsBytes(LOCAL_REPOSITORIES_INFO));
     }
   }
@@ -421,7 +414,7 @@ public class GitHubDataFetcher {
 
     LOGGER.info("Cleaning up local repositories ...");
 
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+    synchronized (LOCAL_REPOSITORIES_INFO) {
       BigInteger total = BigInteger.valueOf(0L);
       for (Map.Entry<URL, LocalRepositoryInfo> entry : LOCAL_REPOSITORIES_INFO.entrySet()) {
         LocalRepositoryInfo info = entry.getValue();
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
@@ -39,6 +39,7 @@ import org.kohsuke.github.HttpException;
 /**
  * Helper class for GitHub data providers which pulls the data from a GitHub repository. Also, the
  * class caches the fetched data for a certain amount of time.
+ * <p>The class is thread-safe.</p>
  */
 public class GitHubDataFetcher {
 
@@ -53,7 +54,7 @@ public class GitHubDataFetcher {
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
   /**
-   * A type reference for serializing {@link #localRepositoriesInfo}.
+   * A type reference for serializing {@link #LOCAL_REPOSITORIES_INFO}.
    */
   private static final TypeReference<HashMap<URL, LocalRepositoryInfo>> LOCAL_REPOSITORIES_TYPE_REF
       = new TypeReference<HashMap<URL, LocalRepositoryInfo>>() {};
@@ -89,6 +90,17 @@ public class GitHubDataFetcher {
   private static final Object LOCAL_REPOSITORIES_INFO_LOCK = new Object();
 
   /**
+   * A cache of local repositories.
+   */
+  static final LRUMap<GitHubProject, LocalRepository> LOCAL_REPOSITORIES
+      = new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE);
+
+  /**
+   * Info about local repositories.
+   */
+  static final Map<URL, LocalRepositoryInfo> LOCAL_REPOSITORIES_INFO = new HashMap<>();
+
+  /**
    * An interface to the GitHub API.
    */
   private GitHub github;
@@ -99,20 +111,9 @@ public class GitHubDataFetcher {
   private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
 
   /**
-   * A cache of local repositories.
-   */
-  final LRUMap<GitHubProject, LocalRepository> localRepositories
-      = new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE);
-
-  /**
    * A base directory.
    */
   private final Path base;
-
-  /**
-   * Info about local repositories.
-   */
-  private final Map<URL, LocalRepositoryInfo> localRepositoriesInfo;
 
   /**
    * Defines how often new updates should be pulled to a local repository.
@@ -166,7 +167,7 @@ public class GitHubDataFetcher {
     this.base = base;
     this.github = github;
 
-    this.localRepositoriesInfo = loadLocalRepositoriesInfo();
+    loadLocalRepositoriesInfo();
   }
 
   /**
@@ -239,14 +240,15 @@ public class GitHubDataFetcher {
    */
   public LocalRepository localRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
+    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+      LocalRepository repository = LOCAL_REPOSITORIES.get(project);
+      if (repository == null) {
+        repository = loadLocalRepositoryFor(project);
+        LOCAL_REPOSITORIES.put(project, repository);
+      }
 
-    LocalRepository repository = localRepositories.get(project);
-    if (repository == null) {
-      repository = loadLocalRepositoryFor(project);
-      localRepositories.put(project, repository);
+      return repository;
     }
-
-    return repository;
   }
 
   /**
@@ -259,46 +261,41 @@ public class GitHubDataFetcher {
   private LocalRepository loadLocalRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
 
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
-      LocalRepositoryInfo info = localRepositoriesInfo.get(project.url());
-      if (info == null) {
-        Path repositoryPath = base
-            .resolve(project.organization().name())
-            .resolve(project.name());
+    LocalRepositoryInfo info = LOCAL_REPOSITORIES_INFO.get(project.url());
+    if (info == null) {
+      Path repositoryPath = base.resolve(project.organization().name()).resolve(project.name());
 
-        info = new LocalRepositoryInfo(
-            repositoryPath, Date.from(Instant.now()), project.url());
-      }
-
-      if (Files.isRegularFile(info.path())) {
-        LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
-        Files.delete(info.path());
-      }
-
-      Optional<Repository> repository = openRepository(info.path());
-      if (!repository.isPresent()) {
-        Files.deleteIfExists(info.path());
-        clone(project, info.path());
-        repository = openRepository(info.path());
-      }
-
-      if (!repository.isPresent()) {
-        throw new IOException("Could not fetch project's repository!");
-      }
-
-      LocalRepository localRepository = new LocalRepository(info, repository.get());
-
-      if (shouldUpdate(localRepository)) {
-        LOGGER.info("Pulling updates from {} ...", project.url());
-        localRepository.pull();
-      }
-
-      info.updated(Date.from(Instant.now()));
-      localRepositoriesInfo.put(project.url(), info);
-      storeLocalRepositoriesInfo();
-
-      return localRepository;
+      info = new LocalRepositoryInfo(repositoryPath, Date.from(Instant.now()), project.url());
     }
+
+    if (Files.isRegularFile(info.path())) {
+      LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
+      Files.delete(info.path());
+    }
+
+    Optional<Repository> repository = openRepository(info.path());
+    if (!repository.isPresent()) {
+      Files.deleteIfExists(info.path());
+      clone(project, info.path());
+      repository = openRepository(info.path());
+    }
+
+    if (!repository.isPresent()) {
+      throw new IOException("Could not fetch project's repository!");
+    }
+
+    LocalRepository localRepository = new LocalRepository(info, repository.get());
+
+    if (shouldUpdate(localRepository)) {
+      LOGGER.info("Pulling updates from {} ...", project.url());
+      localRepository.pull();
+    }
+
+    info.updated(Date.from(Instant.now()));
+    LOCAL_REPOSITORIES_INFO.put(project.url(), info);
+    storeLocalRepositoriesInfo();
+
+    return localRepository;
   }
 
   /**
@@ -384,19 +381,19 @@ public class GitHubDataFetcher {
    * @return The info about local repositories.
    * @throws IOException If something went wrong.
    */
-  protected Map<URL, LocalRepositoryInfo> loadLocalRepositoriesInfo() throws IOException {
+  protected void loadLocalRepositoriesInfo() throws IOException {
     synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
       Path path = base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE);
 
       if (!Files.exists(path)) {
-        return new HashMap<>();
-      }
-
-      if (!Files.isRegularFile(path)) {
+        LOCAL_REPOSITORIES_INFO.clear();
+        return;
+      } else if (!Files.isRegularFile(path)) {
         throw new IOException(String.format("Hey! %s is not a file!", path));
       }
 
-      return MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF);
+      LOCAL_REPOSITORIES_INFO
+          .putAll(MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
     }
   }
 
@@ -409,7 +406,7 @@ public class GitHubDataFetcher {
     synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
       Files.write(
           base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE),
-          MAPPER.writeValueAsBytes(localRepositoriesInfo));
+          MAPPER.writeValueAsBytes(LOCAL_REPOSITORIES_INFO));
     }
   }
 
@@ -417,15 +414,16 @@ public class GitHubDataFetcher {
    * Cleans up local repositories.
    *
    * @param strategy Decides which repositories need to cleaned up.
+   * @throws IOException If something goes wrong.
    */
-  public void cleanup(CleanupStrategy strategy) {
+  public void cleanup(CleanupStrategy strategy) throws IOException {
     Objects.requireNonNull(strategy, "Hey! Cleanup strategy can't be null!");
 
     LOGGER.info("Cleaning up local repositories ...");
 
     synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
       BigInteger total = BigInteger.valueOf(0L);
-      for (Map.Entry<URL, LocalRepositoryInfo> entry : localRepositoriesInfo.entrySet()) {
+      for (Map.Entry<URL, LocalRepositoryInfo> entry : LOCAL_REPOSITORIES_INFO.entrySet()) {
         LocalRepositoryInfo info = entry.getValue();
         if (!Files.exists(info.path())) {
           continue;
@@ -434,7 +432,7 @@ public class GitHubDataFetcher {
       }
 
       Set<URL> toBeRemoved = new HashSet<>();
-      for (Map.Entry<URL, LocalRepositoryInfo> entry : localRepositoriesInfo.entrySet()) {
+      for (Map.Entry<URL, LocalRepositoryInfo> entry : LOCAL_REPOSITORIES_INFO.entrySet()) {
         URL url = entry.getKey();
         LocalRepositoryInfo info = entry.getValue();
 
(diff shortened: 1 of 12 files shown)


sourabhsparkala:
Test cases covered
- `testUpdateGithubInstance()`
  - Update the GitHub instance of `fetcher` and confirm that after update the new `Github` instance is used
- `testLoadAndCleanRepository()`
  - A basic workflow
  - Load a new project
    - Verify if the local repository exists
  - Clean the project
    - Verify that local repository does not exist
  - Confirm that `local_repositories_info.json` is updated after each modification
- `testCleanAndLoadRepository()`
  - A basic workflow
  - Clean a nonexistent project
    - Verify that local repository does not exist
  - Load a new project
    - Verify if the local repository exists
  - Confirm that `local_repositories_info.json` is updated after each modification
- `testMultiGithubInstanceAndAccessRepo()`
  - Create multiple GitHub instances and update using fetcher and confirm if the fetcher is using the latest GitHub instance.
  - Confirm that fetcher is using updated GitHub instance when accessing repo and updating the list
- `testMultiThreadReadAndUpdate()`
  - Multi-Threaded approach
  - Verify `synchronized` block. Initiate a thread (first-thread) to clean an existing project from local repository.
  - During cleanup, initiate another thread (second-thread) to load project from repository
  - The second-thread should wait for the completion of first-thread
- `testMultiThreadLocalRepoInfoTest()` and `testMultiThreadCleanUpTest()`
  - Multi-Threaded approach
  - Parallel regressive execution of multiple threads calling a single synchronized method with a single project and then confirming that there is no disturbance in the final output.
- `testLocalRepositoriesInfoJson()`
  - Load new projects (2)
    - Confirm that `local_repositories_info.json` is updated
  - Clean a project (1)
    - Confirm that `local_repositories_info.json` is updated

artem-smotrakov:
> `localRepositoriesInfo` need not be static as it is read consistently by `loadLocalRepositoriesInfo()` which contains a `synchronized` block in `GitHubDataFetcher`
> * Also, with new `GitHubDataFetcher` instance, this Map is loaded with the current list of local repositories. As expected.

`localRepositoriesInfo` contains info about repositories that have been already cloned. The info is loaded from `DEFAULT_LOCAL_REPOSITORIES_INFO_FILE`. The file reflects the content of `DEFAULT_DIRECTORY`. Since all instances of `GitHubDataFetcher` use the cloned repositories, then they need to share the same instance of `localRepositoriesInfo`. That's why `localRepositoriesInfo` should be static. The access to the info should be synchronized to avoid possible race conditions.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -101,7 +101,7 @@
   /**
    * A cache of local repositories.
    */
-  final LRUMap<GitHubProject, LocalRepository> localRepositories
+  static final LRUMap<GitHubProject, LocalRepository> localRepositories
       = new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE);
```
artem-smotrakov, line 105: @mibo Did you want to be able to set capacity of this cache?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -101,7 +101,7 @@
   /**
    * A cache of local repositories.
    */
-  final LRUMap<GitHubProject, LocalRepository> localRepositories
+  static final LRUMap<GitHubProject, LocalRepository> localRepositories
```
artem-smotrakov, line 104: If you'd like to make it static, please follow the naming conventions for static field. It should be `LOCAL_REPOSITORIES`. The static field should be moved to the other static fields.

artem-smotrakov, line 104: `LRUMap` is not thread-safe.

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html

Please make sure that there are no race conditions when the cache is accessed by multiple threads.

artem-smotrakov:
@sourabhsparkala Could  you Please put a note to the description of the class that says that the class is thread-safe?

```
<p>The class is thread-safe.</p>
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: I have moved the variable usage inside the `synchronized` block to be thread-safe.

mibo:
> > `localRepositoriesInfo` need not be static as it is read consistently by `loadLocalRepositoriesInfo()` which contains a `synchronized` block in `GitHubDataFetcher`
> > 
> > * Also, with new `GitHubDataFetcher` instance, this Map is loaded with the current list of local repositories. As expected.
> 
> `localRepositoriesInfo` contains info about repositories that have been already cloned. The info is loaded from `DEFAULT_LOCAL_REPOSITORIES_INFO_FILE`. The file reflects the content of `DEFAULT_DIRECTORY`. Since all instances of `GitHubDataFetcher` use the cloned repositories, then they need to share the same instance of `localRepositoriesInfo`. That's why `localRepositoriesInfo` should be static. The access to the info should be synchronized to avoid possible race conditions.

Yes, this is the reason why I think the map instance needs to be static.
The list is only loaded during instance creation (in c'tor with `loadLocalRepositoriesInfo()`).
But afterwards it is never re-loaded (only stored e.g. after cleanup) which lead to problems if we have two instances which one runs a clean up and the seconds may have entries in the `localRepositoriesInfo` which are already deleted.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
mibo: No, the cache/map I mentioned was the [entries](https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java#L35) map in the `GitHubDataCache`.
As the default and not changeable size of [5](https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataCache.java#L25) looks a bit small.
With the sixth request (via `GitHubDataFetcher#repositoryFor`) the first cached entry is already removed (if only one instance of the `GitHubDataFetcher` is used).
In case the single `GitHubDataFetcher` is used in multiple threads this may lead to the case that the cached entries are removed before they get requested a second time (which make the cache a bit useless).

Personally I like if such cache sizes are configurable so that whoever use it can adjust it to their needs.

sourabhsparkala: Now the `localRepositories` is also made `static` and its usages are added within `synchronized` block. This should avoid multiple thread clearing the cache unknowingly.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -89,30 +90,30 @@
...
    */
-  private final Path base;
+  private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
 
   /**
-   * Info about local repositories.
+   * A base directory.
    */
-  private final Map<URL, LocalRepositoryInfo> localRepositoriesInfo;
+  private final Path base;
```
artem-smotrakov, line 116: The `base` path is used for loading info about local repositories. A user is allowed to pass any base directory while creating an instance of `GitHubDataFetcher`. It results to re-loading `LOCAL_REPOSITORIES_INFO` that is now static and shared between all instances of the fetcher. As a result, creating a new instance of the class with its own base path leads to unexpected updates in the other instances.

The `base` field should be removed. All instances of the fetcher should use only the default location of local repositories. The info should be initially loaded in a static block, not in a constructor. The methods, which work only with static fields, should be static.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -259,46 +261,41 @@ public LocalRepository localRepositoryFor(GitHubProject project) throws IOExcept
...
-
-        info = new LocalRepositoryInfo(
-            repositoryPath, Date.from(Instant.now()), project.url());
-      }
-
-      if (Files.isRegularFile(info.path())) {
-        LOGGER.warn("{} is a file but it should be a directory, let's remove it", info.path());
-        Files.delete(info.path());
-      }
+    LocalRepositoryInfo info = LOCAL_REPOSITORIES_INFO.get(project.url());
```
artem-smotrakov, line 264: All access to `LOCAL_REPOSITORIES_INFO` should be synchronized. Please don't remove the code for obtaining the lock.

I think that we can just use a monitor of `LOCAL_REPOSITORIES_INFO` to synchronize the access:

```java
synchronized (LOCAL_REPOSITORIES_INFO) {
...
}
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -239,14 +240,15 @@ public GHRepository repositoryFor(GitHubProject project) throws IOException {
    */
   public LocalRepository localRepositoryFor(GitHubProject project) throws IOException {
     Objects.requireNonNull(project, "On no! Project is null!");
+    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
```
artem-smotrakov, line 243: `LOCAL_REPOSITORIES_INFO_LOCK` is used to synchronize access to `LOCAL_REPOSITORIES_INFO`. When you use it to synchronize access to `LOCAL_REPOSITORIES` as well, it means that threads are going to wait for the same lock even if they would like to access different fields. I see that this method calls `loadLocalRepositoryFo()` that accesses `LOCAL_REPOSITORIES_INFO`. However, it happens only when the cache doesn't have an element. I think it would be better to use different locks for those fields. But please make sure that you don't introduce a deadlock.

sourabhsparkala: To avoid this deadlock and synchronization issues. I suggest to replace `LRUMap` and `Map` with `ConcurrentHashMap`. More information can be found [here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html). We would still need `synchronized` blocks while loading the repo and cleanup tasks. But at least the`maps` `localRepositoriesInfo` and `localRepositories` would be thread-safe.

@artem-smotrakov @mibo what do you think?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: I think this would be better. Since `LOCAL_REPOSITORIES_INFO` is `static final`. Only one instance will be available across multiple instances of fetcher. Hence, we can use `synchronized(LOCAL_REPOSITORIES_INFO)`. Instead of having a separate lock object `synchronized(LOCAL_REPOSITORIES_INFO_LOCK)`.

@artem-smotrakov @mibo

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: `LRUMap` is necessary to keep only elements that were recently used. Maybe you could use `Collections.synchronizedMap()` that is mentioned here

https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/LRUMap.html

sourabhsparkala: 👍

sourabhsparkala:
Added 2 more test cases
- `testMultipleInstanceFetcher()`
  - This is verify that the `static` maps are not affected by multiple instances of `GithubDataFetcher`
- `testLocalRepsitoriesCapacity()`
  - To verify the maximum capacity of `LRUMap` `LOCAL_REPOSITORIES`

sourabhsparkala:
@artem-smotrakov I have addressed the review

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -39,6 +39,7 @@
 /**
  * Helper class for GitHub data providers which pulls the data from a GitHub repository. Also, the
  * class caches the fetched data for a certain amount of time.
+ * <p>The class is thread-safe.</p>
```
artem-smotrakov, line 42: Let's please use HTML markup.

```suggestion
 * <p>Helper class for GitHub data providers which pulls the data from a GitHub repository. Also, the
 * class caches the fetched data for a certain amount of time.</p>
 * <p>The class is thread-safe.</p>
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -84,35 +85,34 @@
   private static final boolean SCAN_UNTIL_REMOVABLE = true;
 
   /**
-   * Lock to synchronize access to localRepositoriesInfo.
+   * A synchronized cache of local repositories.
    */
-  private static final Object LOCAL_REPOSITORIES_INFO_LOCK = new Object();
+  static final Map<GitHubProject, LocalRepository> LOCAL_REPOSITORIES = Collections
+      .synchronizedMap(new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE));
```
artem-smotrakov, line 91:
```suggestion
  static final Map<GitHubProject, LocalRepository> LOCAL_REPOSITORIES 
      = Collections.synchronizedMap(
          new LRUMap<>(LOCAL_REPOSITORIES_CACHE_CAPACITY, SCAN_UNTIL_REMOVABLE));
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -84,35 +85,34 @@
...
-   */
-  private GitHub github;
-
-  /**
-   * A limited capacity cache to store the repository of a {@link GitHubProject}.
+   * Synchronized map containing info about local repositories.
    */
-  private final GitHubDataCache<GHRepository> repositoryCache = new GitHubDataCache<>();
+  static final Map<URL, LocalRepositoryInfo> LOCAL_REPOSITORIES_INFO =
+      Collections.synchronizedMap(new HashMap<>());
```
artem-smotrakov, line 97:
```suggestion
  static final Map<URL, LocalRepositoryInfo> LOCAL_REPOSITORIES_INFO 
      = Collections.synchronizedMap(new HashMap<>());
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -384,19 +367,15 @@ protected Date expiration() {
...
-      return MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF);
+  private static void loadLocalRepositoriesInfo() throws IOException {
+    Path path = DEFAULT_DIRECTORY.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE);
+    if (!Files.exists(path)) {
+      LOCAL_REPOSITORIES_INFO.clear();
+    } else if (!Files.isRegularFile(path)) {
+      throw new IOException(String.format("Hey! %s is not a file!", path));
+    } else {
+      LOCAL_REPOSITORIES_INFO
+          .putAll(MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
```
artem-smotrakov, line 378: What is the reason for using if-else block here? If no, please revert it, and leave only the necessary updates.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -405,27 +384,28 @@ protected Date expiration() {
...
    * @throws IOException If something went wrong.
    */
-  protected void storeLocalRepositoriesInfo() throws IOException {
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+  protected static void storeLocalRepositoriesInfo() throws IOException {
+    synchronized (LOCAL_REPOSITORIES_INFO) {
       Files.write(
-          base.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE),
-          MAPPER.writeValueAsBytes(localRepositoriesInfo));
+          DEFAULT_DIRECTORY.resolve(DEFAULT_LOCAL_REPOSITORIES_INFO_FILE),
```
artem-smotrakov, line 390: If you change `DEFAULT_DIRECTORY` to be an instance of `Path`, you can also change the type of `DEFAULT_LOCAL_REPOSITORIES_INFO_FILE` as well.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -384,19 +367,15 @@ protected Date expiration() {
    * @return The info about local repositories.
    * @throws IOException If something went wrong.
    */
-  protected Map<URL, LocalRepositoryInfo> loadLocalRepositoriesInfo() throws IOException {
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
```
artem-smotrakov, line 388: Well, strictly speaking, this method should be synchronized as well. However, since it's only called once during initialization of the class, there should not be any problem. However, if the method may be called multiple times for some reason in the future, there may be problems. If I were you, I would keep synchronization for completeness even if it's not strictly necessary.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -405,27 +384,28 @@ protected Date expiration() {
...
   }
 
   /**
    * Cleans up local repositories.
    *
    * @param strategy Decides which repositories need to cleaned up.
+   * @throws IOException If something goes wrong.
    */
-  public void cleanup(CleanupStrategy strategy) {
+  public void cleanup(CleanupStrategy strategy) throws IOException {
```
artem-smotrakov, line 401: I think this method should also clean `LOCAL_REPOSITORIES` up if a local repository is removed. Could you please address this or create an issue for that?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+    Path project1Dir = projectDir(firstProject.name());
+    GitHub github = mock(GitHub.class);
+    try (Repository repository =
+        FileRepositoryBuilder.create(project1Dir.resolve(".git").toFile())) {
+      repository.create();
+
+      createRepoTestFile(repository, firstProject, "testFile1");
+
+
+      fetcher = spy(new TestGitHubDataFetcher(github));
```
artem-smotrakov, line 157: Is `spy()` required here? Do you use intercept any calls of `TestGitHubDataFetcher`? If so, why don't you update the test class to do what you need?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+
+      fetcher.addRepositoryInfoForTesting(firstProject, project1Dir);
+
+      localRepositoryFor(firstProject, 1);
+      assertEquals(fetcher.github(), github);
```
artem-smotrakov, line 162: I am not sure if I understand the purpose of this test. First, it creates a fetcher with an instance of `GitHub`. Then, it loads a local repository. Finally, it checks if `github()` returns an expected instance of `GitHub`. But loading local repositories doesn't use `GitHub`. The `testUpdateGithubInstance()` already checks that setting an instance of `GitHub` via constrictor works fine.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+        }
+      };
+
+      GitHubProject secondProject = new GitHubProject("test", "project2");
+
+      Runnable loadRepoTask = () -> {
+        try {
+          assertTrue(latch.await(10, TimeUnit.SECONDS));
+          fetcher.localRepositoryFor(secondProject);
+          checkLocalRepository(secondProject, 1);
```
artem-smotrakov, line 233: By using `await()` the test tries to run the cleanup thread before this thread. Why does it need it? What if we remove the `await()` call? I also think that the `checkLocalRepository()` call should be moved to the parent thread.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
sourabhsparkala: The purpose of `cleanup` is to remove the files right? Since `LOCAL_REPOSITORIES` is used as an overall cache. I am not sure if the same purpose applies here. Maybe we should rename this cache?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+        executorService.submit(() -> {
+          try {
+            localRepositoryFor(project);
+          } catch (IOException e) {
+            throw new IllegalStateException("Load repository exception");
+          }
+          latch.countDown();
+        });
+      }
+      latch.await();
```
artem-smotrakov, line 272: I think the test can just use `ExecutinService.awaitTermination()`, can't it?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+      createRepoTestFile(repository, firstProject, "testFile1");
+
+      fetcher = spy(new TestGitHubDataFetcher(github));
+
+      fetcher.addRepositoryInfoForTesting(firstProject, project1Dir);
+
+      localRepositoryFor(firstProject, 1);
+    }
+
+    // Create the second project with second GitHub mock instance.
```
artem-smotrakov, line 335: Is this comment relevant? If I understand correctly, the test checks loading local repositories. But this doesn't involve the `GitHub` instance.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+      latch.await();
+
+      checkCleanUp(project, 0);
+    } finally {
+      executorService.shutdown();
+    }
+  }
+
+  @Test
+  public void testLocalRepositoriesInfoJson() throws IOException {
```
artem-smotrakov, line 317: Does this test check the content of the JSON file?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+
+      localRepositoryFor(secondProject, 2);
+    }
+
+    cleanup(firstProject, 1);
+    cleanup(secondProject, 0);
+  }
+
+  @Test
+  public void testMultipleInstanceFetcher() throws IOException {
```
artem-smotrakov, line 354: This test looks very similar to the previous one. As far as I see, the only difference is that this test doesn't call `cleanup()`. Isn't it a dup? What exactly does it check?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+      fetcher.addRepositoryInfoForTesting(secondProject, project2Dir);
+
+      localRepositoryFor(secondProject, 2);
+    }
+  }
+
+  @Test
+  public void testLocalRepsitoriesCapacity() throws IOException {
+    GitHub github = mock(GitHub.class);
+    fetcher = spy(new TestGitHubDataFetcher(github));
```
artem-smotrakov, line 392: I am not sure why you always wrap it with `spy()`.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+    for (int i = 0; i < 110; i++) {
+      fetcher.addForTesting(mock(GitHubProject.class), mock(LocalRepository.class));
+    }
+    assertEquals(100, fetcher.getLocalRespositoriesForTesting().size());
```
artem-smotrakov, line 396: Would it be better to use `LOCAL_REPOSITORIES_CACHE_CAPACITY`? Would it make sense to check the content of the cache?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +48,89 @@ public void init() throws IOException {
...
 
-    @Override
-    protected void storeLocalRepositoriesInfo() {
-      // do nothing
+    /**
+     * Returns the path of base directory.
+     *  
+     * @return path of type {@link Path}. 
+     */
+    private Path baseDir() {
```
artem-smotrakov, line 132: Let's please remove this method and just use `DEFAULT_DIRECTORY`.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +48,89 @@ public void init() throws IOException {
...
-    protected Map<URL, LocalRepositoryInfo> loadLocalRepositoriesInfo() {
-      return new HashMap<>();
+    /**
+     * Returns a resolved valid path of the project.
+     *  
+     * @param projectName The name of the project.
+     * @return path of type {@link Path}. 
+     */
+    static Path projectDir(String projectName) {
+      return DEFAULT_DIRECTORY.resolve(projectName);
```
artem-smotrakov, line 124: How about the following? I think it would make the code a bit shorter.

```suggestion
    static Path directoryFor(GitHubProject project) {
      return DEFAULT_DIRECTORY.resolve(project.name());
```

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +48,89 @@ public void init() throws IOException {
...
+    }
+
+    /**
+     * Returns the current list of local repositories stored in the base folder.
+     *
+     * @return Map of local repositories available.
+     * @throws IOException if something goes wrong.
+     */
+    public Map<URL, LocalRepositoryInfo> getLocalRespositoriesInfoForTesting() throws IOException {
+      return LOCAL_REPOSITORIES_INFO;
```
artem-smotrakov, line 97: `LOCAL_REPOSITORIES_INFO` has package-private access. Is this method really necessary? I see that only tests from the same package needs this field.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +48,89 @@ public void init() throws IOException {
...
-      localRepositories.put(project, repository);
+      LOCAL_REPOSITORIES.put(project, repository);
+    }
+
+    /**
+     * Returns the current list of local repositories stored in the cache.
+     *
+     * @return Map of local repositories in cache.
+     */
+    public Map<GitHubProject, LocalRepository> getLocalRespositoriesForTesting() {
```
artem-smotrakov, line 86: `LOCAL_REPOSITORIES` has package-private access. Is this method really necessary? I see that only tests from the same package needs this field.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
artem-smotrakov: > The purpose of `cleanup` is to remove the files right?

Right. If a local repository is removed, then the elements in `LOCAL_REPOSITORIES`, which point to the removed repository, should be remove as well. It's not about renaming. `LOCAL_REPOSITORIES` may contain invalid items if the files have been removed.

artem-smotrakov: I can open an issue for that. Let me know.

sourabhsparkala: I will try to address it. If not will create an issue.

sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/251

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +92,349 @@ public void testShouldUpdate() throws IOException {
...
+
+  @Test
+  public void testLoadAndCleanRepository() throws IOException, InterruptedException {
+    GitHubProject project = new GitHubProject("test", "project");
+    Path projectDir = projectDir(project.name());
+    try (
+        Repository repository = FileRepositoryBuilder.create(projectDir.resolve(".git").toFile())) {
+      repository.create();
+
+      createRepoTestFile(repository, project, "testfile");
```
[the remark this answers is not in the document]
sourabhsparkala: Not required

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: You are right. This test actually doesn't test anything new. Earlier I had written the test case with the mindset that github plays a crucial role in gathering information about local repositories. But we are not testing that part.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Updated in https://github.com/SAP/fosstars-rating-core/pull/246#discussion_r448757291

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: `ExecutionService.awaitTermination()` takes in `timeout` argument. Which would be hard termination.
I wanted to make sure that all the threads are executed at least once.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: The entire test is another duplicate. I am testing the `json` file is by calling `loadLocalRepositoriesInfo()` in `getLocalRespositoriesInfoForTesting()`(class `TestGitHubDataFetcher`). Hence, since it is being checked in every other test. It does not need to be verified explicitly here.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Updated here https://github.com/SAP/fosstars-rating-core/pull/246#discussion_r448768601

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Scenario:
- I wanted to confirm that even after creating multiple instances of `GitHubDataFetcher`, the associated static variable like `LOCAL_REPOSITORIES_INFO` is not over-written or cleared. 
- To confirm that the static variable will still contain information even after the `GitHubDataFetcher` instance changes.
- I understand I am basically checking static functionality. 

Let me know if this is still not required. I will remove it.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Not required

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: yes

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
sourabhsparkala: I have updated this method to 
```
loadLocalRepositoriesInfo();
return LOCAL_REPOSITORIES_INFO;
```

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: The scenario :
- First thread is running the `cleanup` task.
- Second thread is running the `loadLocalRepsotories` task.
We wanted to pause the first thread for some time. To confirm that the second thread will not cross the synchronized block until the first thread is complete. Hence, this approach. 

-`You are right `checkLocalRepository()` should be checked in the main thread. Have done this change

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Using this check to confirm if all the threads have completed the task with `assertTrue(latch.await(5, TimeUnit.SECONDS));`

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: https://github.com/SAP/fosstars-rating-core/pull/246#discussion_r448773776

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +90,227 @@ public void testShouldUpdate() throws IOException {
...
+  }
+
+  @Test
+  public void testLocalRepsitoriesCapacity() throws IOException {
+    GitHub github = mock(GitHub.class);
+    fetcher = new TestGitHubDataFetcher(github);
+    for (int i = 0; i < 110; i++) {
+      fetcher.addForTesting(mock(GitHubProject.class), mock(LocalRepository.class));
+    }
+    assertEquals(LOCAL_REPOSITORIES_CACHE_CAPACITY, LOCAL_REPOSITORIES.size());
```
artem-smotrakov, line 283: Again, does it make sense to make sure that the content of the cache is correct? Or not?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
artem-smotrakov: I can't find anymore the test called the `cleanup()` method in the end. Looks like you just removed it. I think that was a good test.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +90,227 @@ public void testShouldUpdate() throws IOException {
...
+    try (Repository repository =
+        FileRepositoryBuilder.create(project1Dir.resolve(".git").toFile())) {
+      repository.create();
+
+      fetcher = new TestGitHubDataFetcher(github);
+      fetcher.addRepositoryInfoForTesting(firstProject, project1Dir);
+      testLocalRepositoryFor(firstProject, 1);
+    }
+
+    // Create the second project with second GitHub mock instance.
```
artem-smotrakov, line 263: Is this comment still relevant? I don't see the test creates another `GitHub`.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
artem-smotrakov: It's not about renaming. I mean that the method looks redundant. I think you can just remove it and use `LOCAL_REPOSITORIES_INFO` in the tests from the same package.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
artem-smotrakov: > We wanted to pause the first thread for some time. To confirm that the second thread will not cross the synchronized block until the first thread is complete. Hence, this approach.

I am not sure why we need it. The most interesting part here is to let the threads run as they run.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
artem-smotrakov: If the threads didn't finish, then `await()` return false, and `assertTrue()` throws an exception.

If the threads didn't finish, then `ExecutinService.awaitTermination()` would throw an exception.

In both cases, the tests should fail.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -384,19 +373,20 @@ protected Date expiration() {
...
+        return;
       }
 
       if (!Files.isRegularFile(path)) {
         throw new IOException(String.format("Hey! %s is not a file!", path));
       }
 
-      return MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF);
+      LOCAL_REPOSITORIES_INFO
+          .putAll(MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
```
artem-smotrakov, line 389:
```suggestion
      LOCAL_REPOSITORIES_INFO.putAll(
          MAPPER.readValue(Files.newInputStream(path), LOCAL_REPOSITORIES_TYPE_REF));
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcher.java
```diff
@@ -405,27 +395,28 @@ protected Date expiration() {
    *
    * @throws IOException If something went wrong.
    */
-  protected void storeLocalRepositoriesInfo() throws IOException {
-    synchronized (LOCAL_REPOSITORIES_INFO_LOCK) {
+  protected static void storeLocalRepositoriesInfo() throws IOException {
```
artem-smotrakov, line 398:
```suggestion
  private static void storeLocalRepositoriesInfo() throws IOException {
```

Or, package-private.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -79,4 +90,227 @@ public void testShouldUpdate() throws IOException {
...
+      repository.create();
+
+      fetcher.addRepositoryInfoForTesting(project, projectDir);
+      testLocalRepositoryFor(project, 1);
+      testCleanup(project, 0);
+    }
+  }
+
+  @Test
+  public void testCleanAndLoadRepository() throws IOException, InterruptedException {
```
artem-smotrakov, line 121: Does it throw InterruptedException?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +49,72 @@ public void init() throws IOException {
...
     }
 
+    /**
+     * Adds {@link GitHubProject} and its associated {@link LocalRepository} details to cache.
+     * @param project The {@link GitHubProject}.
+     * @param repository The {@link LocalRepository}.
+     */
     public void addForTesting(GitHubProject project, LocalRepository repository) {
-      localRepositories.put(project, repository);
+      LOCAL_REPOSITORIES.put(project, repository);
```
artem-smotrakov, line 79: The method should be static.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
```diff
@@ -53,42 +49,72 @@ public void init() throws IOException {
...
-    protected Map<URL, LocalRepositoryInfo> loadLocalRepositoriesInfo() {
-      return new HashMap<>();
+    
+    /**
+     * Add new project to be considered while loading repository.
+     *  
+     * @param project The {@link GitHubProject}.
+     * @param projectDir The local {@link Path} for the {@link GitHubProject}.
+     */
+    public void addRepositoryInfoForTesting(GitHubProject project, Path projectDir) {
```
artem-smotrakov, line 99: The method should be static.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: Yes, it does. I have implemented that.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/TestGitHubDataFetcherHolder.java
sourabhsparkala: I have removed this redundant method definition

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: It does not. Removing that.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
sourabhsparkala: It was this method which had `cleanup()`. As I mentioned all the tests are basically reloading from the `.json` file before the check. Hence, this explicit and duplicate test for the same did not make sense.

Referring to https://github.com/SAP/fosstars-rating-core/pull/246#discussion_r448865394

artem-smotrakov:
@sourabhsparkala I have made the following updates:

- Fixed formatting issues.
- Simplified the code.
- Fixed a couple of bugs in the tests.
- Added a system property to set the base directory for repositories.

Please have a look and let me know if it's okay.

sourabhsparkala:
Looks good @artem-smotrakov . Thank you for the update.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/GitHubDataFetcherTest.java
```diff
@@ -71,12 +88,245 @@ public void testShouldUpdate() throws IOException {
...
+      executorService.execute(loadRepoTask);
+    } finally {
+      executorService.awaitTermination(4, TimeUnit.SECONDS);
+    }
+
+    checkLocalRepository(secondProject, 1);
+  }
+
+  @Test
+  public void testMultiThreadLocalRepoInfoTest() throws IOException, InterruptedException {
```
sourabhsparkala, line 188: Is it possible to just rename this test method? I overshot the name here.

artem-smotrakov: If you have a better name, yes, go ahead and rename it.

sourabhsparkala:
@artem-smotrakov let me know if its good to merge now. Thanks for the review.

Issue #232, opened 2020-06-19
New feature for Maven GPG plugin


artem-smotrakov:
This plugin signs all of the project's attached artifacts with GnuPG. The security awareness score can check if a project uses this plugin to sign artifacts.

http://maven.apache.org/plugins/maven-gpg-plugin/sign-mojo.html

Here is an example:

https://github.com/apache/dubbo/blob/master/pom.xml#L378

Here is a list of things to do:

- [x] Add a new feature.
- [x] Implement a data provider that fills out the new feature.
- [x] Include the new feature to the open-source security rating.
- [x] Update test and test vectors.
- [x] Update the demo tool if necessary (enable the provider, update the pretty printer, etc)

PR #247, opened 2020-06-30
Data provider should check the entire POM file

### LINKED ISSUE #234: Data providers should check the entire pom.xml
The report for Apache Flink says that it doesn't use OWASP Dependency Check

https://github.com/SAP/fosstars-rating-core/blob/master/docs/oss/security/apache/flink.md - it says No but the project uses the plugin

but the project actually uses it:

https://github.com/apache/flink/blob/master/pom.xml#L2015

Looks like the data providers can't find it for some reason. Maybe they ignore pluginManagement section and profiles.

artem-smotrakov:

- Updated the existing data providers to check profiles and plugin management sections in POM files.
- Updated the tests.

This closes #234


sourabhsparkala:
However, after the discussion regarding the possible introduction of a Helper class which can be used by all the Data Providers when parsing `pom.xml` plugins and configurations.

artem-smotrakov:
> However, after the discussion lets introduce a Helper class that can be used by all the Data Providers when parsing `pom.xml` plugins and configurations.

@sourabhsparkala I've implemented visitors for POM files. Please have a look.

src/main/java/com/sap/sgs/phosphor/fosstars/maven/AbstractModelVisitor.java
```diff
@@ -0,0 +1,21 @@
+package com.sap.sgs.phosphor.fosstars.maven;
+
+import java.util.Set;
+import org.apache.maven.model.Plugin;
+import org.apache.maven.model.ReportPlugin;
+
+/**
+ * An implementation of {@link ModelVisitor} that doesn't nothing.
```
sourabhsparkala, line 8: typo

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyCheck.java
```diff
@@ -213,4 +117,36 @@ private static boolean isDependencyCheck(ReportPlugin plugin) {
...
+
+  /**
+   * Creates a new visitor for searching for OWASP Dependency Check in a POM file.
+   */
+  private static Visitor withVisitor() {
+    return new Visitor();
+  }
+
+  /**
+   * A visitor for searching for OWASP Dependency Check in a POM file.
```
sourabhsparkala, line 129: A visitor for searching OWASP Dependency Check?

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpToolTest.java
```diff
@@ -58,14 +56,11 @@ public void testGradleWithoutNoHTTP() throws IOException {
...
-    when(content.read()).thenReturn(is);
+    final LocalRepository repository = mock(LocalRepository.class);
+    when(repository.read(filename)).thenReturn(Optional.of(is));
 
-    GHRepository repository = mock(GHRepository.class);
-    when(repository.getFileContent(filename)).thenReturn(content);
-
-    when(fetcher.github().getRepository(any())).thenReturn(repository);
+    GitHubProject project = new GitHubProject("org", "test");
+    fetcher.addForTesting(project, repository);
```
sourabhsparkala, line 63: Is there no test case to check where `LOCATION` it was expected to be found?

sourabhsparkala:
@artem-smotrakov minor comments and questions.

src/test/java/com/sap/sgs/phosphor/fosstars/data/github/UsesNoHttpToolTest.java
artem-smotrakov: Currently there is no tests for that. In fact, locations are not used at all here. They are necessary for #232. The test will be added there. However, I can also add tests for the visitors that use the existing POM files.

artem-smotrakov:
I've added a test. Thanks for the review and comments @sourabhsparkala !

PR #255, opened 2020-07-03
WeightedCompositeScore should set weight for N/A sub-scores

### LINKED ISSUE #254: WeightedCompositeScore should set weight for N/A sub-scores
Currently `WeightedCompositeScore` skips setting weights for sub-scores that are not applicable.

artem-smotrakov:
This fixes #254


artem-smotrakov:
There was an issue with the test. I fixed that. The patch for the score remains the same.

artem-smotrakov:
Thanks for the review @mibo !

PR #258, opened 2020-07-07
Add a label for unclear ratings

### LINKED ISSUE #253: Add a label for unclear ratings
To better indicate that the calculated rating has a low confidence the label should be `UNCLEAR` instead of the calculated (and mostly `BAD` label).

It should be possible to set a threshold to define all calculated ratings with a confidence below the threshold will have the new `UNCLEAR` label.

Like it was done for the [security statistics overview](https://github.com/SAP/fosstars-rating-core/tree/be5711a9757341663a1c63df6ba5f6535b1c351c/docs/oss/security).

artem-smotrakov:

- Added a new label `OssSecurityRating.SecurityLabel.UNCLEAR`.
- Updated `OssSecurityRating.label()` method to return the new label.
- Updated the demo tool to print out the new label.
- Updated `SecurityRatingAnalysis.ipynb` notebook to consider the new label.

This fixes #253


src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
```diff
@@ -55,12 +57,15 @@ public OssSecurityScore score() {
...
-  /**
-   * Implements a mapping from a score to a label.
-   */
   @Override
-  protected SecurityLabel label(double score) {
-    Score.check(score);
+  protected SecurityLabel label(ScoreValue scoreValue) {
+    Objects.requireNonNull(scoreValue, "Oh no! Score value is null!");
+
+    if (scoreValue.confidence() < thresholds.ubclear) {
```
sourabhsparkala, line 64: Typo.

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
```diff
@@ -81,30 +86,39 @@ protected SecurityLabel label(double score) {
     /**
      * The default thresholds.
      */
-    public static final Thresholds DEFAULT = new Thresholds(5.0, 8.0);
+    public static final Thresholds DEFAULT = new Thresholds(5.0, 8.0, 8.0);
```
sourabhsparkala, line 89: Is it possible that unclear threshold may be too high? Is Confidence usually `>8.0`?

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
artem-smotrakov: Good catch!

src/main/java/com/sap/sgs/phosphor/fosstars/model/rating/oss/OssSecurityRating.java
artem-smotrakov: I even think it could be higher. The rating is currently based on 26 features. If the threshold is 0.8, than it is okay if approximately up to 5 features are unknown (well, it actually depends on weights).

Confidence for most of projects in the report is more than 9:

https://github.com/SAP/fosstars-rating-core/blob/master/src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb

sourabhsparkala: ok 👍

PR #263, opened 2020-07-08
Introduce new features for breaking build by OWASP Dependency Scan

### LINKED ISSUE #237: New feature for breaking build by OWASP Dependency Check
OWASP Dependency Check can break build if vulnerabilities are found

https://github.com/apache/druid/blob/master/pom.xml#L1542

It should positively affect the score for dependency scans. The score can also consider other info from the plugin config such as:

- CVSS threshold
- The place where the tool is called (the default build section, reporting section, profile)
- and so on

Here is a list of things to do

- [ ] Add new features.
- [ ] Update the existing data provider to fill out the new features.
- [ ] Introduce a new score for OWASP Dependency Check.
- [ ] Add tests and test vectors for the new score.

sourabhsparkala:

- Introduced new features and a score that shows how the project uses UsesOwaspDependencyScan.
- Added a new data provider that gathers information about the new features.
- Introduced new tests and test vectors.

This fixes #237


src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +217,53 @@ private OssFeatures() {
...
+   */
+  public static final Feature<Boolean> USES_OWASP_DEPENDENCY_CHECK_AS_OPTIONAL_SCAN =
+      new BooleanFeature("If a project uses OWASP Dependency Check scan on specific occasions");
+
+  /**
+   * If OWASP Dependency plugin is used maybe to report vulnerabilities during the build.
+   */
+  public static final Feature<Boolean> USES_OWASP_DEPENDENCY_CHECK_AS_OPTIONAL_REPORT =
+      new BooleanFeature(
+          "If a project uses OWASP Dependency Check to generate report on specific occasions");
```
artem-smotrakov, line 244: Some combinations of those feature are invalid. For example, `USES_OWASP_DEPENDENCY_CHECK_AS_MANDATORY_SCAN == true and USES_OWASP_DEPENDENCY_CHECK_AS_OPTIONAL_SCAN == true` is invalid. The scores, which uses those features, must check for such invalid configurations. That would make them more complex. I would recommend to re-design the features to eliminate such invalid configurations. For example, you can add a feature that describes how the tool is used in the project. The feature can take three values: not used, mandatory or optional. If the tool runs, it always generates a report.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +217,53 @@ private OssFeatures() {
...
+      new DoubleFeature(
+          "OWASP Dependency check CVSS Threshold score used to fail the build");
+
+  /**
+   * Check if configuration to fail the build exists in OWASP Dependency. Identify whether the build
+   * fails if atleast one vulnerability is found.
+   */
+  public static final Feature<Boolean> USES_OWASP_DEPENDENCY_CHECK_FAIL_ANY_VULNERABILITY =
+      new BooleanFeature(
+          "If a project uses OWASP Dependency check to fail the build on finding vulnerability");
```
artem-smotrakov, line 268: Do we really need a separate feature for each parameter? In fact, is it allowed to use them together? Aren't they alternatives? I think they can be replaced by one double feature with a range [0, 11].

sourabhsparkala: Actually this is a good idea, if the 
- junitFailOnCVSS - then CVSS score is already provided
- failBuildOnCVSS - then CVSS score is already provided
- failBuildOnAnyVulnerability - then CVSS score is 0, as it considers all the vulnerabilities.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
sourabhsparkala: Yes having 3 outcomes sounds better.
I will do the changes accordingly.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
artem-smotrakov: > Actually this is a good idea, if the
> 
> * junitFailOnCVSS - then CVSS score is already provided
> * failBuildOnCVSS - then CVSS score is already provided
> * failBuildOnAnyVulnerability - then CVSS score is 0, as it considers all the vulnerabilities.

Yes, sounds good to me. And if none of the parameters are provided, then CVSS is set to 11 which means that the tool doesn't fail if any issue is found. It looks a bit ugly though. Another way is to add a boolean feature that is set to true if no threshold provided.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
    */
   public static final Feature<Boolean> SIGNS_ARTIFACTS
       = new BooleanFeature("If a project signs artifacts");
+
+  /**
+   * If OWASP Dependency plugin is used to scan the project. It is either used as a mandatory step,
```
artem-smotrakov, line 224: The tools is called OWASP Dependency Check. It's not required to use it via plugins. In fact, there are multipel plugins for the tool. Please update the description.
```suggestion
   * Shows if OWASP Dependency Check is used to scan a project. It is either used as a mandatory step,
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
...
+   * optional step or not used at all.
+   */
+  public static final Feature<Status> OWASP_DEPENDENCY_CHECK_SCAN_AVAILABILITY =
+      new AvailabilityFeature("OWASP Dependency Check scan mode");
```
artem-smotrakov, line 228: At the moment, I am not sure if we're going to use `Status` and `AvailabilityFeature` in other places. Therefore let's please make them specific. Please rename `Status` to `OwaspDependencyCheckUsage`. Then, let's please rename `AvailabilityFeature` to `OwaspDependencyCheckUsageFeature`. I know it is quite a long name and looks a bit ugly but I believe it has to be specific here. Finally, please rename `AvailabilityValue` to `OwaspDependencyCheckUsageValue` accordingly.

In fact, we have the `EnumValue` class. But unfortunately Jackson fails to deserialize it. This disabled test shows reproduces the issue

https://github.com/SAP/fosstars-rating-core/blob/master/src/test/java/com/sap/sgs/phosphor/fosstars/model/value/EnumValueTest.java#L88

I am not sure if newer versions of Jackson work fine but https://github.com/FasterXML/jackson-databind/issues/2605 seems to be fixed. So, you can try to use the `EnumValue` instead of introducing `OwaspDependencyCheckUsageValue`.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
...
+
+  /**
+   * Check if configuration to fail the build exists in OWASP Dependency. Identify the CVSS
+   * threshold score to fail the build on finding vulnerabilities greater than the threshold.
```
artem-smotrakov, line 232:
```suggestion
   * Shows if OWASP Dependency Check is configured to fail the build if vulnerabilities are found. The feature contains a threshold for CVSS score that is used by the tool.
   *
   * @see <a href="https://jeremylong.github.io/DependencyCheck/">OWASP Dependency Check documentation</a>
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
...
+   */
+  public static final Feature<Double> OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD =
+      new DoubleFeature(
```
artem-smotrakov, line 235: Would you mind introducing a new `BoundedDoubleFeature`? Like the following:

https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/BoundedIntegerFeature.java

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
    */
   public static final Feature<Boolean> SIGNS_ARTIFACTS
       = new BooleanFeature("If a project signs artifacts");
+
+  /**
+   * If OWASP Dependency plugin is used to scan the project. It is either used as a mandatory step,
+   * optional step or not used at all.
+   */
+  public static final Feature<Status> OWASP_DEPENDENCY_CHECK_SCAN_AVAILABILITY =
```
artem-smotrakov, line 227:
```suggestion
  public static final Feature<Status> OWASP_DEPENDENCY_CHECK_USAGE =
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
artem-smotrakov, line 228:
```suggestion
      new AvailabilityFeature("How OWASP Dependency Check is used");
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -216,4 +219,19 @@ private OssFeatures() {
...
+      new AvailabilityFeature("OWASP Dependency Check scan mode");
+
+  /**
+   * Check if configuration to fail the build exists in OWASP Dependency. Identify the CVSS
+   * threshold score to fail the build on finding vulnerabilities greater than the threshold.
+   */
+  public static final Feature<Double> OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD =
+      new DoubleFeature(
+          "OWASP Dependency check CVSS Threshold score used to fail the build");
```
artem-smotrakov, line 236:
```suggestion
          "A CVSS threshold for OWASP Dependency Check to fail the build");
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Status.java
```diff
@@ -0,0 +1,8 @@
+package com.sap.sgs.phosphor.fosstars.model.value;
+
+/**
+ * The enum availability status.
+ */
+public enum Status {
+  MANDATORY, OPTIONAL, NOTFOUND
```
artem-smotrakov, line 7:
```suggestion
  MANDATORY, OPTIONAL, NOT_USED
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,318 @@
...
+   * 
+   * @param value the value to be parsed.
+   * @param type the type of the string value.
+   * @return the parsed value.
+   */
+  private static double parseScore(String value, String type) {
+    if ("number".equals(type)) {
+      return Double.parseDouble(value);
+    }
+    return 0.0;
```
artem-smotrakov, line 244: To prevent potential issues, it would be better to return 0 if the the type is `boolean` and throw an exception in case of an unexpected type.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,318 @@
...
+    Xpp3Dom dom = ((Xpp3Dom) configuration).getChild(configName);
+    if (dom == null) {
+      return Optional.empty();
+    }
+
+    return Optional.ofNullable(dom.getValue());
+  }
+
+  /**
+   * Returns whether the plugin will be executed as a mandatory step during the build.
```
artem-smotrakov, line 268:
```suggestion
   * Checks whether the plugin will be executed as a mandatory step during the build.
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,318 @@
...
+   * 
+   * @param defaultParam The location of the plugin.
+   * @param locations The set of {@link Location}.
+   * @return True if the plugin will be executed as a mandatory step, false otherwise.
+   */
+  private static boolean mandatoryCheckFor(Set<Location> locations) {
+    return (locations.contains(BUILD) || locations.contains(REPORTING))  
+        && !locations.contains(PROFILE)
+        && !locations.contains(MANAGEMENT);
```
artem-smotrakov, line 277: If the plugin is mentioned in both the default build section and plugin management, then the method returns false which seems to be wrong.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,318 @@
...
+    /**
+     * The availability status of plugin in the project.
+     */
+    private Status status = NOTFOUND;
+
+    @Override
+    public void accept(Plugin plugin, Set<Location> locations) {
+      if (isDependencyCheck(plugin)) {
+        status = availablityFrom(locations);
+        score = cvssScoreFrom(plugin.getConfiguration());
```
artem-smotrakov, line 306: Did you check if the tool can still fail the build even if it's in the reporting section? If it can't, then the provide needs to take this into account when it reads the CVSS threshold.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,135 @@
...
+import com.sap.sgs.phosphor.fosstars.model.qa.TestVectors;
+import com.sap.sgs.phosphor.fosstars.model.score.FeatureBasedScore;
+import com.sap.sgs.phosphor.fosstars.model.value.ScoreValue;
+import com.sap.sgs.phosphor.fosstars.model.value.Status;
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.commons.lang3.Range;
+
+/**
+ * The scores assesses how well an open-source project uses OWASP dependency check to scan
```
artem-smotrakov, line 22:
```suggestion
 * The scores assesses how well an open-source project uses OWASP Dependency Check to scan
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,135 @@
...
+   * @see <a href="https://nvd.nist.gov/vuln-metrics/cvss">CVSS</a> 
+   */
+  private static final double MAX_CVSS_SCORE = 10.0;
+  
+  /**
+   * The minimum CVSS score.
+   * 
+   * @see <a href="https://nvd.nist.gov/vuln-metrics/cvss">CVSS</a> 
+   */
+  private static final double MIN_CVSS_SCORE = 0.0;
```
artem-smotrakov, line 43: Please use the existing constants

https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/sgs/phosphor/fosstars/model/value/CVSS.java

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,135 @@
...
+  
+  /**
+   * Initializes a new {@link OwaspDependencyScanScore}.
+   */
+  OwaspDependencyScanScore() {
+    super("How a project uses Owasp Dependency to scan dependencies for vulnerabilities",
```
artem-smotrakov, line 49:
```suggestion
    super("How a project uses Owasp Dependency Check to scan dependencies for vulnerabilities",
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,135 @@
...
+  @Override
+  public ScoreValue calculate(Value... values) {
+    Value<Status> availabilityValue =
+        find(OWASP_DEPENDENCY_CHECK_SCAN_AVAILABILITY, values);
+    Value<Double> cvssScore =
+        find(OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD, values);
+
+    ScoreValue scoreValue = scoreValue(Score.MIN, availabilityValue, cvssScore);
+
+    // If the project does not use OWASP Dependency scans.
```
artem-smotrakov, line 63:
```suggestion
    // if the project does not use OWASP Dependency Check
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,135 @@
...
+      return scoreValue;
+    }
+
+    // If the project uses OWASP Dependency scans to identify vulnerable dependencies, then we are
+    // happy.
+    if (availability.equals(MANDATORY)) {
+      scoreValue.set(7);
+    } else if (availability.equals(OPTIONAL)) {
+      scoreValue.set(3);
+    }
```
artem-smotrakov, line 75: Could you please define constants for 7 and 3?

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
artem-smotrakov: The threshold belongs to the interval [0, 10] and it can be also not set at all. How about defining a special feature for that? Let's set `OwaspDependencyCheckCvssThreshold`.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
sourabhsparkala: Not sure why but `EnumValue` still fails. Updated `jackson-databind to 2.11.1`

`com.fasterxml.jackson.databind.exc.InvalidDefinitionException:
Cannot construct instance of `java.lang.Enum` (no Creators, like default construct, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: (BufferedInputStream); line: 38, column: 12] (through reference chain: com.sap.sgs.phosphor.fosstars.model.qa.TestVectors["elements"]->java.util.ArrayList[1]->com.sap.sgs.phosphor.fosstars.model.qa.StandardTestVector["values"]->java.util.HashSet[0]->com.sap.sgs.phosphor.fosstars.model.value.EnumValue["value"])`

artem-smotrakov: Maybe the issue still exists, maybe some annotation need to be added. Anyway. we can work on in it later.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
sourabhsparkala: Is it better to have this `BoundedDoubleFeature` alternative to already existing `BoundedIntegerFeature`?

artem-smotrakov: I meant something different but I am not sure that I have a clear understanding how it can be implemented. Feel free to add a new `BoundedDoubleFeature` class. It would be also good to define a constant for the `11` score in single place. However, I am not sure where it should be.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,318 @@
...
+    Optional<InputStream> content = repository.read("build.gradle");
+
+    if (content.isPresent()) {
+      try (BufferedReader reader = new BufferedReader(new InputStreamReader(content.get()))) {
+        String line;
+        while ((line = reader.readLine()) != null) {
+          if (line.trim().contains("org.owasp:dependency-check-gradle")) {
+            return new ValueHashSet(
+                OWASP_DEPENDENCY_CHECK_SCAN_AVAILABILITY.value(MANDATORY),
+                OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD.value(11.0));
```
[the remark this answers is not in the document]
sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/272

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/oss/OssFeatures.java
sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/273

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/ValueHashSet.java
```diff
@@ -77,6 +77,18 @@ public ValueHashSet(Value... values) {
     }
   }
 
+  /**
+   * Initializes a {@link ValueHashSet} with a number of {@link ValueSet}.
+   *
+   * @param valueSets The sets of values.
+   */
+  public ValueHashSet(ValueSet... valueSets) {
```
artem-smotrakov, line 85: Is it still used anywhere?

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,127 @@
...
+
+    // if the project fails build based on certain CVSS threshold. If CVSS of a vulnerability is
+    // greater than threshold, then the build fails. Lower the threshold score means most
+    // vulnerabilities are considered, then we are happy.
+    double failThreshold = cvssScore.orElse(OwaspDependencyCheckCvssThreshold.OUT_OF_BOUND);
+    if (failThreshold == CVSS.MIN) {
+      scoreValue.increase(STEP_SCORE);
+    } else if (inBetween(failThreshold)) {
+      scoreValue.increase(CVSS.MAX / (STEP_SCORE * failThreshold));
+    }
```
artem-smotrakov, line 77: How about just using a linear function for the threshold?

```java
if (CVSS.isValid(failThreshold)) {
  scoveValue.increase(STEP_SCORE * (CVSS.MAX - failThreshold) / CVSS.MAX);
}
```

Or, if you add `OwaspDependencyCheckCvssThresholdValue`:

```java
if (cvssScore.isSpecified()) {
  scoveValue.increase(STEP_SCORE * (CVSS.MAX - cvssScore.get()) / CVSS.MAX);
}
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,301 @@
...
+    private OwaspDependencyCheckUsage status = NOT_USED;
+
+    /**
+     * The flag which closes down when a plugin location is found.
+     */
+    private boolean flag = true;
+
+    @Override
+    public void accept(Plugin plugin, Set<Location> locations) {
+      if (isDependencyCheck(plugin) && flag) {
```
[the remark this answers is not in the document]
sourabhsparkala: I have put in a test case to verify that [here](https://github.com/SAP/fosstars-rating-core/blob/ec22cddd318f900867648596ac00bb98b06d4433/src/test/resources/com/sap/sgs/phosphor/fosstars/data/github/MavenWithOwaspDependencyCheckInBuildAndProfile.xml).

Also, the `MavenUtils#browse()` function does not check in the `profiles` first. The precedence is always given to `build` section.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/ValueHashSet.java
sourabhsparkala: No this is not used anymore

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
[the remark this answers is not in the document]
artem-smotrakov: We should not rely on the order implemented in `browse()`. What if the order changes? Then, we should remember to update this date provider. Let's please not rely on the browsing order.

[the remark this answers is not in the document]
sourabhsparkala: I resolved this.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+    checkMaven(repository, values);
+    checkGradle(repository, values);
+
+    return values;
+  }
+
+  /**
+   * Returns the set of OWASP Dependency Check features.
+   */
+  static Set<Feature> features() {
```
artem-smotrakov, line 85: I am not sure if this method and `OWASP_FEATURES` are necessary. We already have `supportedFeatures()` that should return a set of supported features. Let's keep it simple unless there is a strong reason.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+ * features:
+ * <ul>
+ *   <li>{@link OssFeatures#OWASP_DEPENDENCY_CHECK_USAGE}</li>
+ *   <li>{@link OssFeatures#OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD}</li>
+ * </ul>
+ */
+public class UsesOwaspDependencyScan extends GitHubCachingDataProvider {
+
+  /**
+   * A list of configurations associated to OWASP Dependency check plugin which can fail the build.
```
artem-smotrakov, line 46:
```suggestion
   * A list of configurations associated to OWASP Dependency Check plugin which can fail the build.
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+        OWASP_DEPENDENCY_CHECK_USAGE, 
+        OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD);
+  }
+
+  /**
+   * Checks if a project uses OWASP Dependency Check Maven plugin.
+   *
+   * @param repository The project's repository.
+   * @param values set of type {@link ValueSet}.
+   * @return True if the project uses the plugin, false otherwise.
```
artem-smotrakov, line 96: The method doesn't return anything.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+   * @param value the value to be parsed.
+   * @param type the type of the string value.
+   * @return the parsed value.
+   */
+  private static double parseScore(String value, String type) {
+    switch (type) {
+      case "number":
+        return Double.parseDouble(value);
+      case "boolean":
+        return 0.0;
```
artem-smotrakov, line 224: If `failBuildOnAnyVulnerability` is explicitly set to `false`, should the method return `0.0`?

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+
+  /**
+   * Get the value if a configuration exists.
+   * 
+   * @param name The name of the configuration to find.
+   * @param configuration The list of configurations associated to OWASP Dependency Check plugin.
+   * @return The value of the configuration if found.
+   */
+  private static Optional<String> parameter(String name, Object configuration) {
+    if (configuration == null || !(configuration instanceof Xpp3Dom)) {
```
artem-smotrakov, line 238: I think the null-check if redundant here

```suggestion
    if (configuration instanceof Xpp3Dom == false) {
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/OwaspDependencyCheckUsageFeature.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.sap.sgs.phosphor.fosstars.model.Value;
+import com.sap.sgs.phosphor.fosstars.model.value.OwaspDependencyCheckUsage;
+import com.sap.sgs.phosphor.fosstars.model.value.OwaspDependencyCheckUsageValue;
+import java.util.Objects;
+
+/**
+ * A feature which holds a {@link OwaspDependencyCheckUsage} value. Usage status mandatory,
+ * optional or not found.
```
artem-smotrakov, line 11: Maybe it's going to have more or less elements. Therefore, the javadoc would need to be updated. Let's just drop it.

```suggestion
 * A feature which holds a {@link OwaspDependencyCheckUsage} value.
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,128 @@
...
+
+    ScoreValue scoreValue = scoreValue(MIN, usageValue, cvssScore);
+
+    // if the project does not use OWASP Dependency Check
+    OwaspDependencyCheckUsage usage = usageValue.orElse(NOT_USED);
+    if (usage.equals(NOT_USED)) {
+      return scoreValue;
+    }
+
+    // if the project uses OWASP Dependency scans to identify vulnerable dependencies, then we are
```
artem-smotrakov, line 61:
```suggestion
    // if the project uses OWASP Dependency Check to identify vulnerable dependencies, then we are
```

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
```diff
@@ -0,0 +1,128 @@
...
+        OWASP_DEPENDENCY_CHECK_USAGE,
+        OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD);
+  }
+
+  @Override
+  public ScoreValue calculate(Value... values) {
+    Value<OwaspDependencyCheckUsage> usageValue =
+        find(OWASP_DEPENDENCY_CHECK_USAGE, values);
+    Value<Double> cvssScore =
+        find(OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD, values);
```
artem-smotrakov, line 51: Should it simplify a bit the code below?

```suggestion
    OwaspDependencyCheckUsageValue usageValue =
        find(OWASP_DEPENDENCY_CHECK_USAGE, values);
    OwaspDependencyCheckCvssThresholdValue cvssScore =
        find(OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD, values);
```

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
sourabhsparkala: Yes I think it should. If `0.0`, then any CVSS score should be considered, right? 
Or is there something else I may be missing?

artem-smotrakov: I think if `failBuildOnAnyVulnerability` is false, then the threshold is not specified (unless it's set in another option).

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+   * @return The cvss score which indicates how the build should fail. Otherwise null.
+   */
+  private static Double cvssScoreFrom(Object configuration) {
+    // The OWASP Dependency Check plugin is found and checking configurations.
+    for (String name : FAIL_BUILD_CONFIGURATIONS.keySet()) {
+      Optional<String> content = parameter(name, configuration);
+
+      if (content.isPresent()) {
+        String type = FAIL_BUILD_CONFIGURATIONS.get(name);
+        return parseScore(content.get(), type);
```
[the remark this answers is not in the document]
sourabhsparkala: Yes, you right. It is first come first serve basis. I have rearranged it in a way like this

```
FAIL_BUILD_CONFIGURATIONS.put("failBuildOnAnyVulnerability", "boolean");
FAIL_BUILD_CONFIGURATIONS.put("failBuildOnCVSS", "number");
FAIL_BUILD_CONFIGURATIONS.put("junitFailOnCVSS", "number"); 
```

I want to give preference to the first configuration, as the score is assumed to be `0.0` -> The best case.
Similarly the next 2.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
sourabhsparkala: Can this be written like this then?

`!(configuration instanceof Xpp3Dom)`

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
```diff
@@ -0,0 +1,292 @@
...
+  private static void update(ValueSet values, Map<OwaspDependencyCheckUsage, Double> usageMap) {
+    OwaspDependencyCheckUsage usage = NOT_USED;
+
+    if (usageMap.containsKey(MANDATORY)) {
+      usage = MANDATORY;
+    } else if (usageMap.containsKey(OPTIONAL)) {
+      usage = OPTIONAL;
+    }
+
+    values.update(OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD.value(usageMap.get(usage)));
```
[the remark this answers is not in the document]
sourabhsparkala: Yes, I have a test for this.

I assumed this would happen as `usageMap.put()` is used only if a plugin is found. Hence, this way.

```
if (isDependencyCheck(plugin)) {
        usageMap.put(usageFrom(locations), cvssScoreFrom(plugin.getConfiguration()));
 }
``` 
`OWASP_DEPENDENCY_CHECK_FAIL_CVSS_THRESHOLD.value()` will not throw NPE as, I am doing the null check before I set the value in `OwaspDependencyCheckCvssThreshold#value(Double n)`. Here, if `n` is `null`, then set the `JUNK_VALUE`.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
artem-smotrakov: Sure, it can. I just don't like `(!(`. But not a big deal. Feel free to write like this.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
sourabhsparkala: I tried this way. But IDE suggests that we have to use casts to achieve this. Hence, felt this would be simpler.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
sourabhsparkala: You are right. I have removed this.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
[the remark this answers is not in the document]
artem-smotrakov: I see. I am not sure it's worth it to have this check in `value()`. Please see my comments below.

src/main/java/com/sap/sgs/phosphor/fosstars/model/score/oss/OwaspDependencyScanScore.java
artem-smotrakov: I appreciate IDE's suggestions. If we update the type of `OWASP_DEPENDENCY_CHECK_USAGE` to `OwaspDependencyCheckUsage`, can it make it happy?

artem-smotrakov: Furthermore, if `usageValue`is not `OwaspDependencyCheckUsageValue`, doesn't it look like that something is wrong?

sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/278

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesOwaspDependencyScan.java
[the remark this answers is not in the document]
sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/279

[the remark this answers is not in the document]
artem-smotrakov: I am not sure that I understand the description of the issue you created. Could you please update it?

I didn't mean to introduce a new feature. Here are several specific questions:

- How the data provider should fill out the features if both `failBuildOnAnyVulnerability` and `failBuildOnCVSS` are present?
- How the data provider should fill out the features if both `failBuildOnAnyVulnerability` and `junitFailOnCVSS` are present?
- How the data provider should fill out the features if both `failBuildOnCVSS` and `junitFailOnCVSS` are present?

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/OwaspDependencyCheckCvssThreshold.java
```diff
@@ -0,0 +1,48 @@
...
+ * CVSS threshold is out of range, it would not be considered during score calculation.</p>
+ * 
+ * @see <a href=
+ *      "https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html">
+ *      OWASP Dependency Check configuration</a>
+ */
+public class OwaspDependencyCheckCvssThreshold extends BoundedDoubleFeature {
+
+  /**
+   * A junk value, x < 0.0 or x > 10.0.
```
artem-smotrakov, line 21: `x < 0.0 or x > 10.0` is not true. x is 11. The comment needs to be updated.

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/DoubleValue.java
```diff
@@ -14,7 +14,7 @@
   /**
    * A double value.
    */
-  private final double number;
```
artem-smotrakov, line 17: I am not sure if the access level should be updated. At least, it should be package-private. Or, others can just use the `value()` method.

src/main/java/com/sap/sgs/phosphor/fosstars/model/feature/BoundedDoubleFeature.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Checks if an double belongs to the interval.
+   *
+   * @param n The double to be checked.
+   * @return n after the check. Otherwise throws an {@link IllegalArgumentException}.
+   */
+  private Double check(Double n) {
+    if (Double.compare(n, from) < 0 || Double.compare(n, to) > 0) {
+      throw new IllegalArgumentException(
+          String.format("Feature value has to be in [%d, %d] but %d given!", from, to, n));
```
artem-smotrakov, line 90: Wrong data format specifiers.

artem-smotrakov:
I've rebased the branch and fixed several formatting and minor issues. I've also left a few TODOs in the code. There are still the following open issues (I created several new ones):

https://github.com/SAP/fosstars-rating-core/issues/279
https://github.com/SAP/fosstars-rating-core/issues/282
https://github.com/SAP/fosstars-rating-core/issues/283
https://github.com/SAP/fosstars-rating-core/issues/272
https://github.com/SAP/fosstars-rating-core/issues/284

Once we fix them, we can include the new score to the open-source security rating.

PR #274, opened 2020-07-15
Update weights in the open-source security score

### LINKED ISSUE #269: Update weights in the open-source security score
The weights in the open-source security score were adjusted long time ago. Since that, the sub-scores have changes quite a lot. Therefore, it makes sense to check if the current weights still make sense. If not, they need to be adjusted.

artem-smotrakov:

- Tuned weights in the open-source security score.
- Updated the existing test vectors.

This closes #269


artem-smotrakov:

PR #289, opened 2020-07-22
Check for scores in FeatureBasedScore

### LINKED ISSUE #261: Check for scores in FeatureBasedScore
The `FeatureBasedScore` is supposed to work only with pure features. However, the `Score` interface extends the `Feature` interface. Currently, it's possible to use scores in the `FeatureBasedScore`. That should not be allowed.

artem-smotrakov:
This fixes #261

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/FeatureBasedScore.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/score/FeatureBasedScore.java
@@ -7,6 +7,7 @@ import com.sap.sgs.phosphor.fosstars.model.Feature;
 import com.sap.sgs.phosphor.fosstars.model.Score;
 import com.sap.sgs.phosphor.fosstars.model.Visitor;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/score/FeatureBasedScore.java
```diff
@@ -50,13 +51,18 @@ public FeatureBasedScore(String name, String description, Feature... features) {
...
+      }
+    }
+    this.features = new HashSet<>(features);
   }
 
   @Override
   @JsonIgnore
   public final Set<Feature> features() {
-    return features;
+    return new HashSet<>(features);
```
sourabhsparkala, line 65: The features is already set to `new HashSet<>()` in constructor L59.
We can just return the `features` right?

artem-smotrakov: If the method returns the `features` field, then the caller can modify the field. If the constructor doesn't wrap feature into a new set, then the caller can modify the internal `features` field as well. That's why wrapping is necessary in both places.

PR #319, opened 2020-08-13
Update OwaspSecurityLibraries to support Gradle

### LINKED ISSUE #297: Update OwaspSecurityLibraries to support Gradle
The data provide is currently able to work with Maven projects. It needs to be enhanced to support Gradle as well.

### LINKED ISSUE #318: Make sure that streams are closed
There are several places where i/o streams are not closed. Let's fix them.

artem-smotrakov:

- Updated the `OwaspSecurityLibraries` provider to search for features in Gradle files. This closes #297
- Updated several classes to close input streams. This fixes #318 
- Added new test cases.

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/OwaspSecurityLibraries.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/OwaspSecurityLibraries.java
@@ -64,37 +64,19 @@ public class OwaspSecurityLibraries extends GitHubCachingDataProvider {
   protected ValueSet fetchValuesFor(GitHubProject project) throws IOException {
     logger.info("Figuring out if the project uses OWASP security libraries ...");
 
-    LocalRepository repository = GitHubDataFetcher.localRepositoryFor(project);
-
     ValueSet values = new ValueHashSet();
 
-    checkMaven(repository, values);
-    if (found(values)) {
-      return values;
-    }
-
-    checkGradle(repository, values);
-    if (found(values)) {
-      return values;
-    }
-
+    // set default values
     values.update(USES_OWASP_ESAPI.value(false));
     values.update(USES_OWASP_JAVA_ENCODER.value(false));
     values.update(USES_OWASP_JAVA_HTML_SANITIZER.value(false));
 
-    return values;
-  }
+    LocalRepository repository = GitHubDataFetcher.localRepositoryFor(project);
 
-  /**
-   * Checks if a value set contains the features for OWASP security tools.
-   *
-   * @param values The value set to be checked.
-   * @return True if the value set contains the features, false otherwise.
-   */
-  private static boolean found(ValueSet values) {
-    return values.has(USES_OWASP_ESAPI)
-        && values.has(USES_OWASP_JAVA_ENCODER)
-        && values.has(USES_OWASP_JAVA_HTML_SANITIZER);
+    checkMaven(repository, values);
+    checkGradle(repository, values);
+
+    return values;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/OwaspSecurityLibraries.java
```diff
@@ -56,13 +65,45 @@ protected ValueSet fetchValuesFor(GitHubProject project) throws IOException {
...
 
+  /**
+   * Checks if a value set contains the features for OWASP security tools.
+   *
+   * @param values The value set to be checked.
+   * @return True if the value set contains the features, false otherwise.
+   */
+  private static boolean found(ValueSet values) {
+    return values.has(USES_OWASP_ESAPI)
+        && values.has(USES_OWASP_JAVA_ENCODER)
```
sourabhsparkala, line 96: Should this be connected via `&&`. Does it mean that all the features are inter-dependent?

artem-smotrakov: Yes, the method check if all the features are set.

> Does it mean that all the features are inter-dependent?

I am not sure if I got it. The features are independent.

artem-smotrakov:
> I see that `Files.newInputStream()` is still being used in an unsafe manner in these files:
> 
> * `LocalRepository#file()` 

The `file(Path)` method uses try-with-resources. Please let me know what the issue is.

> and `LocalRepository#read()`

The `LocalRepository.read()` methods are supposed to return a stream. The caller is suppose to read and close it. If the method returns to the caller a closed stream, then the caller won't be able to read the data. I've checked the callers of the method, and they close the streams. Please let me know if you see any caller that doesn't close a stream.

> * `NVD#open()`

The same story.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/OwaspSecurityLibraries.java
sourabhsparkala: Is there a possibility that only one of the features may contain a value. But I guess not, all the features will be assigned a default value.

artem-smotrakov:
> Does search for `Files.walk` also apply here to close the stream. I think I missed the one in `FuzzedInOssFuzz`
> 
> * I see it being used in `LocalRepository` as well.

Yes, according to the docs, the streams should be closed

https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walk-java.nio.file.Path-java.nio.file.FileVisitOption...-

I've updated the methods that use `wald()` and don't close streams.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/OwaspSecurityLibraries.java
artem-smotrakov: The provider should assign a value to all features. If one of the features don't have a value after checking Maven files, then the provide tries to read Gradle files. I've noticed that `checkMaven()` method always sets all values, therefore `checkGradle()` can't be reached because `found()` just uses `ValueSet.has()`. Maybe you meant this problem. That's definitely an issue. I've fixed that. Thanks for paying attention @sourabhsparkala !

sourabhsparkala: Yes, I meant somewhere along the lines. But not exactly as clear as you have explained. This may be a bug in `OwaspDependencyScan` DataProvider as well?

artem-smotrakov: Yes, we can consider updating this provider. Could you please open an issue?

sourabhsparkala: https://github.com/SAP/fosstars-rating-core/issues/323

PR #327, opened 2020-08-20
Added a score for Dependabot

### LINKED ISSUE #326: Add a score for Dependabot
Let's extract a score for Dependabot from `DependencyScanScore`. That's a preparation for #298

artem-smotrakov:
Here is a list of updates:

- Added `DependabotScore` that is extracted from `DependencyScanScore`.
- Added a test and test vectors.

This fixes #326


sourabhsparkala (approved):
Looks good to me 👍

PR #332, opened 2020-08-27
Added a Markdown formatter for the tool

### LINKED ISSUE #320: Create a Markdown formatter for the tool
The report for well-known open-source projects contains an output of `PrettyPrinter`. See for example the following:

https://github.com/SAP/fosstars-rating-core/blob/master/docs/oss/security/GoogleChrome/lighthouse.md

`PrettyPrinter` was originally supposed to be used in the command line tool only. But then we just applied it to generate the report for well-known projects. It's time to create a Markdown formatter. The new report should look nice in GitHub pages.

artem-smotrakov:

- Added a new class `MarkdownFormatter` that builds a Markdown report for a rating.
- Added a new class `CommonFormatter` that provides common methods for the existing formatters.
- Updated the tool to use the new formatter when it generates reports for multiple projects.
- Added a smoke test for the new formatter.
- Updated the report for curl.

Closes #320

### DIFF
--- a/docs/oss/security/curl/curl.md
+++ b/docs/oss/security/curl/curl.md
@@ -1,184 +1,313 @@
 # curl/curl
 
-```
-Here is how the rating was calculated:
-  Score:........Security of project
-  Value:........6.56 out of 10.0
-  Confidence:...High (10.0 out of 10.0)
-  Based on:.....7 sub-scores:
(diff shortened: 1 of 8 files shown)


sourabhsparkala (approved):
Looks good to me 👍

artem-smotrakov:

Issue #336, opened 2020-09-01
Update the report and re-calculate the thresholds (1.0.0)


artem-smotrakov:
Let's update the report for well-known open-source projects, and then re-calculate the thresholds for the labels in the open-source security rating.

PR #337, opened 2020-09-01
Updated the list of projects for the report

artem-smotrakov:
Updated the list of projects for which we calculate security ratings and the thresholds for the labels in the security score:

- Added AWS projects
- Added Google projects
- Added Bouncy Castle projects
- Consider only projects that have more than 1000 stars

Part of #336


artem-smotrakov:
Hearing no objections.

PR #354, opened 2020-11-12
Better UsesGithubForDevelopment provider

### LINKED ISSUE #348: Better double comparison in UsesGithubForDevelopment
The data provider `UsesGithubForDevelopment` does a number of checks for a repository and then return true if enough checks have passed. Here is what it looks like now:

```java
int points = CHECKS.stream()
        .map(check -> check.test(repository) ? 1 : 0)
        .reduce(0, Integer::sum);

    return (double) points / CHECKS.size() >= threshold;
```

The threshold is currently 0.6 and the provider runs 5 checks. It means that each passed check adds 0.2. Three passed checks are 0.6. However, `>=` can still return false in this case. That would lead to a wrong value of the feature. It's better to use the method `Double.compare()` and/or set the threshold to 0.5.

Maybe the threshold value needs to be updated to make the provider less strict.

artem-smotrakov:
Here is a list of updates:

- The provider now returns true if pull requests are enabled
- Decreased the threshold
- Updated the test

Fixes #348

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
@@ -12,6 +12,7 @@ import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.kohsuke.github.GHContent;
 import org.kohsuke.github.GHRepository;
@@ -25,7 +26,7 @@ public class UsesGithubForDevelopment extends CachedSingleFeatureGitHubDataProvi
    * This threshold shows how many checks have to be passed to say that a project uses GitHub
    * for development.
    */
-  private static final double CONFIDENCE_THRESHOLD = 0.6;
+  private static final double CONFIDENCE_THRESHOLD = 0.5;
 
   /**
    * A list of checks to figure out if a project uses GitHub for development.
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -118,26 +114,43 @@ static boolean usesGitHubForDevelopment(GHRepository repository, double threshol
...
 
   /**
    * Check if a string is not a URL to GitHub.
    * 
    * @param input The string to be checked.
    * @return True if the string is not a URL to GitHub, false otherwise.
-   * @throws IOException If something goes wrong.
    */
-  private static boolean notGitHubUrl(String input) throws IOException {
+  static boolean notGitHubUrl(String input) {
```
Sachpat, line 149: Just a question, any specific reason for checking the negativity here? I mean it could have been isGitHubUrl() right?

artem-smotrakov: It could be `isGitHubUrl()`. One of the reasons may be that we are interested in non-GitHub URLs. Maybe @sourabhsparkala can give another reason since he wrote this originally.

src/main/java/com/sap/sgs/phosphor/fosstars/data/github/UsesGithubForDevelopment.java
```diff
@@ -38,17 +39,18 @@
         return StringUtils.isEmpty(description) || !description.toLowerCase().contains("mirror");
       },
 
+      // check if the repository doesn't have an explicit link to the original repository
+      // or an explicit link to an SVN repository and the link does not contain github.com
+      repository -> Stream.of(repository.getMirrorUrl(), repository.getSvnUrl())
+          .filter(StringUtils::isNotEmpty)
+          .noneMatch(UsesGithubForDevelopment::notGitHubUrl),
```
Sachpat, line 46: Please correct me if I am wrong, .noneMatch(UsesGithubForDevelopment::notGitHubUrl) would return true if all of the urls are github right? Because, noneMatch() returns true if none of the stream elements match the given predicate, it returns false if any of the stream elements matches the condition specified by the predicate.

Sachpat: And if yes, this conflicts the above comment 
// or an explicit link to an SVN repository and the link does not contain github.com
right?

artem-smotrakov: The check works as you described. I don't think that the comment conflicts with the check but I find the comment quite quite difficult to read. Therefore, I updated the comment. Hope it's clearer now.

Sachpat (approved):
Looks good to me.

artem-smotrakov:
> Looks good to me.

:+1: Thanks for the review @Sachpat

PR #357, opened 2020-11-18
WeightedCompositeScore should return unknown score if all values are unknown

### LINKED ISSUE #356: WeightedCompositeScore should return unknown score if all values are unknown
Currently, `WeightedCompositeScore` only checks if all subscores are not applicable. It should do the same for unknown values.

artem-smotrakov:
Here is a list of updates:

- Updated `WeightedCompositeScore` to check if all sub-scores are unknown and return an unknown score value if so.
- Test vectors now supports unknown scores.
- Minor improvements.

Fixes #356

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
@@ -16,6 +16,11 @@ public abstract class AbstractTestVector implements TestVector {
   final Interval expectedScore;
 
   /**
+   * If it's set to true, then an unknown score value is expected.
+   */
+  final boolean expectedUnknownScore;
+
+  /**
    * If it's set to true, then a not-applicable score value is expected.
    */
   final boolean expectedNotApplicableScore;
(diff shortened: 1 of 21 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
```diff
@@ -93,14 +108,16 @@ public boolean equals(Object o) {
       return false;
     }
     AbstractTestVector that = (AbstractTestVector) o;
-    return expectedNotApplicableScore == that.expectedNotApplicableScore
+    return expectedUnknownScore == that.expectedUnknownScore
```
sourabhsparkala, line 111: I am not sure, but would `Boolean.compare()` make more sense?

sourabhsparkala (approved):
Looks good . Minor questions 👍

src/main/java/com/sap/sgs/phosphor/fosstars/model/qa/AbstractTestVector.java
artem-smotrakov: `Boolean.compare()` returns an integer. If we used this method, the code would look like `Boolean.compare(expectedUnknownScore, that.expectedUnknownScore) == 0`. I don't think it's worth it.

PR #366, opened 2020-11-24
Updated CodeQL and LGTM scores to take languages into account

### LINKED ISSUE #361: CodeQL and LGTM scores can consider languages
CodeQL supports only a limited set of languages. Therefore, the CodeQL score should return an N/A value for the languages that are not supported.

artem-smotrakov:
Here is a list of updates:

- Updated `CodeqlScore` and `LgtmScore` to take into account programming languages that are used in a project.

Fixes #361

### DIFF
--- a/src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Language.java
+++ b/src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Language.java
@@ -15,7 +15,7 @@ public enum Language {
   // .NET languages
   C_SHARP, F_SHARP, VISUALBASIC,
 
-  PHP, RUBY, PYTHON, JAVASCRIPT,
+  PHP, RUBY, PYTHON, JAVASCRIPT, TYPESCRIPT, GO,
 
   OTHER;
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Language.java
```diff
@@ -40,6 +40,9 @@ public static Language parse(String string) {
         return C_SHARP;
       case "F#":
         return F_SHARP;
+      case "go":
```
mibo, line 43: As `string.toUpperCase()` is used the cases should be `GO` and `GOLANG`.

mibo (changes requested):

src/main/java/com/sap/sgs/phosphor/fosstars/model/value/Language.java
artem-smotrakov: Good catch! I've fixed that.

mibo (approved):

artem-smotrakov:
Thanks for the review @mibo

PR #376, opened 2020-11-27
Moved test vectors to src/test

### LINKED ISSUE #375: Move test vectors to src/test
Currently, all test vectors for the ratings and score are in `src/main` but they are used only in tests. Therefore, it makes sense to move them to `src/test` so that they are not included in the resulting jar file.

artem-smotrakov:
Only refactoring, no new functionality:

- Moved the existing test vectors to `src/test` directory.
- Removed classes for score and rating verification.
- Removed verification tests for scores.
- Added a single test that runs verification for all scores.
- Optimized imports.

Fixes #375


mibo (approved):
Have not checked the deleted files.
But what I have seen for the existing looks good 👍

src/test/java/com/sap/sgs/phosphor/fosstars/model/score/oss/ScoreVerificationTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    boolean failed = false;
+    for (Score score : collector.scores) {
+      String className = score.getClass().getSimpleName();
+      String testVectorsFileName = String.format("%sTestVectors.yml", className);
+      try (InputStream is = score.getClass().getResourceAsStream(testVectorsFileName)) {
+        TestVectors testVectors = TestVectors.loadFromYaml(is);
+        if (testVectors.isEmpty()) {
+          failed = true;
+          System.out.printf("No test vectors for %s%n", className);
```
Sachpat, line 38: DO we really need these sysouts?

artem-smotrakov: Yes, those may be quite helpful when you add a new score and write a verification procedure for it.

PR #396, opened 2021-01-13
Use com.sap.oss.phosphor package prefix

### LINKED ISSUE #393: Use com.sap.oss.phosphor package prefix
The jars are published under `com.sap.oss.phosphor` group id on Maven Central

https://mvnrepository.com/artifact/com.sap.oss.phosphor/fosstars-rating-core

But the `pom.xml` file uses `com.sap.phosphor`

https://github.com/SAP/fosstars-rating-core/blob/a01cb06fb6f2dbaf271d4a9ed861a5b9af923c15/pom.xml#L7

Furthermore, the classes also use `com.sap.phosphor` in the package name.

Let's update group id and package names to use `com.sap.oss.phosphor`.

- [x] Update `pom.xml`
- [x] Updates package names in the classes
- [x] Updated the README and docs

artem-smotrakov:
Renamed `com.sap.sgs.phosphor` package to `com.sap.oss.phosphor`.

Closes #393


mibo:
A bit annoying but I completely agree that the group id and package names should be aligned 👍

mibo (approved):

artem-smotrakov:
> A bit annoying but I completely agree that the group id and package names should be aligned

Yeah, I don't like this update either.

Issue #401, opened 2021-01-21
PoC: GitHub action for creating a report for a project


artem-smotrakov:
Let's try to add a GitHub action that creates/updates a report for the project where it's run. The action can be based on the command-line tool.

PR #403, opened 2021-01-21
The command-line tool should store a report in a file

### LINKED ISSUE #402: The command-line tool should store a report in a file
Currently, the command-line tool prints everything out to stdout. It would be good if the tool had a parameter that tells it to store a report in a file. Another option can specify the format of the report. Furthermore, it would be nice to have an option that tells the tool to store a raw rating in JSON.

artem-smotrakov:
Added new options for the command-line tool: `--report-file`, `--report-type`, `--raw-rating-file`.

The new options are necessary for #401

Fixes #402

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
@@ -18,12 +18,15 @@ import com.sap.oss.phosphor.fosstars.tool.InputString;
 import com.sap.oss.phosphor.fosstars.tool.Reporter;
 import com.sap.oss.phosphor.fosstars.tool.YesNoQuestion;
 import com.sap.oss.phosphor.fosstars.tool.YesNoQuestion.Answer;
+import com.sap.oss.phosphor.fosstars.tool.format.Formatter;
+import com.sap.oss.phosphor.fosstars.tool.format.MarkdownFormatter;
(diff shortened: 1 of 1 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
```diff
@@ -240,18 +260,45 @@ void run() throws IOException {
...
+   */
+  private static void checkOptionsIn(CommandLine commandLine) {
+    if (!commandLine.hasOption("url") && !commandLine.hasOption("config")
+        && !commandLine.hasOption("gav")) {
+
+      throw new IllegalArgumentException(
+          "You have to give me either --url, --gav or --config option!");
+    }
+
+    if (commandLine.hasOption("report-type") && !commandLine.hasOption("report-file")) {
```
mibo, line 277: Will approve the PR but I recommend to extract the parameter names/re-used strings like this to constants.

mibo (approved):
Minor suggestion but overall looks fine

src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SecurityRatingCalculator.java
artem-smotrakov: I know that such constants may make life easier if we'd like to change its value. But I find it more readable if we have the values here instead of constant names. The parameters of the command-line tool don't change often.

Issue #399, opened 2021-01-19
Provide the command-line tool as a standalone application


artem-smotrakov:
It may be possible to use Java Packager to build packages for Windows, Mac and Linux

https://docs.oracle.com/javase/10/tools/javapackager.htm#JSWOR719

Issue #398, opened 2021-01-19
Run the command-line tool in a Docker container


artem-smotrakov:
If the command-line tool is available in a Docker container, then a user won't need to build it by itself. It would be nice if the Docket image in deployed to the Docket Hub (or at least to the GitHub Docker storage).

PR #404, opened 2021-01-21
PoC: GitHub action for creating a report for a project

### LINKED ISSUE #401: PoC: GitHub action for creating a report for a project

artem-smotrakov:

- Added a local GitHub action that calculates a security rating for the current project. See README for the action for more details.
- Added a workflow that runs the action on this repository.
- Added a badge with security rating to the main README.

Closes #401


mibo (approved):
LGTM

mibo:
I approved it as it looks good and works.

However I have one idea for an improvement.
As I understand it the GH action runs the Dockerfile (`fosstars-create-single-report/Dockerfile`).
In this Dockerfile the `fosstars-rating-core` project is always checked out and build.
Would it not better to either create a Docker container which already have the `fosstars-rating-core` pre-build?
For each stable `fosstars-rating-core` a container could be provided.
Another option would to just download the released CLI jar file (e.g. from Maven central or GitHub releases).

artem-smotrakov:
> I approved it as it looks good and works.

Thanks for the review @mibo

> In this Dockerfile the `fosstars-rating-core` project is always checked out and build.
> Would it not better to either create a Docker container which already have the `fosstars-rating-core` pre-build?

Sounds like a good idea. I think that would definitely speed it up. I am not sure if it's going to work though. The guideline tells us to create a Dockerfile that is built when the action runs

https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action

> For each stable `fosstars-rating-core` a container could be provided.

I am not sure if we can use it for the action but I am planning to provide a Docker image with the command-line tool and publish it on Docker Hub, please see https://github.com/SAP/fosstars-rating-core/issues/398

> Another option would to just download the released CLI jar file (e.g. from Maven central or GitHub releases).

Yes, this way, we could avoid building Fosstars in the Dockerfile. Unfortunately, it's not currently possible since we don't deploy the jar with the command-line tool to Maven Central. Once we do, we can update the Dockerfile.

I'm also planning to create platform-specific packages with the command-line tool, please see https://github.com/SAP/fosstars-rating-core/issues/399. Once it's there, we can use `deb` or `rpm` package in the Dockerfile.

mibo:
> Sounds like a good idea. I think that would definitely speed it up. I am not sure if it's going to work though. The guideline tells us to create a Dockerfile that is built when the action runs
> 
> https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action
> 
> I am not sure if we can get rid of the Dockerfile. But we can probably speed it up. We can also use multistage builds.

The idea is to use another base image in the Dockerfile which already contains the CLI (jar)...

> I am not sure if we can use it for the action but I am planning to provide a Docker image with the command-line tool and publish it on Docker Hub, please see #398
 
...what exactly is what you planned with #398  🙂  (what I have missed).

> I'm also planning to create platform-specific packages with the command-line tool, please see #399. Once it's there, we can use `deb` or `rpm` package in the Dockerfile.

Cool ideas 👍

artem-smotrakov:
> The idea is to use another base image in the Dockerfile which already contains the CLI (jar)...

Yeah, I think we can use a Docker image which already contains the CLI once we have such an image.

Feel free to open issues if you have other ideas.

PR #406, opened 2021-01-27
Set default locale to en/US for tests

### LINKED ISSUE #407: PrettyPrinterTest fails if not-English locale is set

mibo:
Set default locale to en/US for tests as the test fails on a machine with another (not en) locale.

Currently the locale is only set but not reset after the test.
As from my understanding resetting is not required.
But please let me know if it is required to set it back to previous locale for some reason.

Closes #407


artem-smotrakov:
> Currently the locale is only set but not reset after the test.
> As from my understanding resetting is not required.
> But please let me know if it is required to set it back to previous locale for some reason.

I am not sure whether it's going to affect the tests that run after this one. If they run in the same JVM (I think they do), then they other tests will see the new locale. However, that won't necessary make them fail. However, I'd prefer to restore the original locale to be on the safe side and just to be consistent. I've updated the patch.

artem-smotrakov (approved):

PR #417, opened 2021-02-08
Postpone downloading NVD in the command-line tool

### LINKED ISSUE #400: Postpone downloading NVD in the command-line tool
The command-line tools starts downloading NVD before validating all input parameters. As a result, it a user runs the tool with invalid input, then it's going to be discovered after downloading NVD. But it can fail earlier without wasting time for downloading NVD.

artem-smotrakov:

 - Updated the `NVD` class to download data from NVD if necessary.
- Updated the command-line tool not to download NVD in advance.

Closes #400

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/nvd/NVD.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/nvd/NVD.java
@@ -26,7 +26,8 @@ import org.apache.logging.log4j.Logger;
 import us.springett.nistdatamirror.NistDataMirror;
 
 /**
- * This class offers an interface to the NVD.
+ * <p>This class offers an interface to the NVD.</p>
+ * <p>The class is not thread-safe.</p>
  */
 public class NVD {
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/nvd/NVD.java
```diff
@@ -24,7 +26,8 @@
 import us.springett.nistdatamirror.NistDataMirror;
 
 /**
- * This class offers an interface to the NVD.
+ * <p>This class offers an interface to the NVD.</p>
+ * <p>The class is not threat-safe.</p>
```
mibo, line 30:
```suggestion
 * <p>The class is not thread-safe.</p>
```

mibo (approved):
Only a minor typo, hence I approved it already 🙂

src/main/java/com/sap/oss/phosphor/fosstars/nvd/NVD.java
artem-smotrakov: My usual typo! Thanks!

PR #425, opened 2021-02-11
Improved CodeqlProvider and command-line tool

### LINKED ISSUE #405: CodeqlProvider doesn't detect all configs
Looks like `CodeqlProvider` doesn't always recognize CodeQL jobs. To reproduce it, run the command-line tool on Fosstars repository. It uses CodeQL but the tool doesn't detect it.

### LINKED ISSUE #424: Command-line tool doesn't always detect log4j config
https://github.com/SAP/fosstars-rating-core/blob/b4af1c4bc35b6824e42ca9bca4b4c0fca32d9266/src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SecurityRatingCalculator.java#L220

When the tool runs in the directory where Fosstars was build, the config is loaded. But when it runs it another directory, it looks like it can't load the config. As a result, no output is printed out.

artem-smotrakov:

- Updated `CodeqlProvider` to read config files directly instead of delegating it to `LocalRepository.read()`. This method resolves input path against the repository path. But sometimes the input path may be absolute. As a result, read fails.
- Use the standard file name for Log4j config, and update log level at runtime if `--verbose` is set.
- Updated tests.

Fixes #405
Fixes #424


mibo (approved):
LGTM

PR #426, opened 2021-02-11
Make ValueSet iterable

### LINKED ISSUE #418: Make ValueSet iterable
`ValueSet` should implement the `Iterable` interface

https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html

That would make it possible to use a `ValueSet` in a for-each loop. The `toArray()` method may be then removed.

artem-smotrakov:
Here is a list of updates:

- Removed `ValueSet.toArray()` method.
- Added `ValueSet.toSet()` method.
- Updated `ValueSet` to implement `Iterable`.
- Fixed warnings for Java generics.
- Updated tests.

Fixes #418


mibo (approved):
LGTM 👍

artem-smotrakov:
Thanks for the review @mibo!

PR #433, opened 2021-02-18
Improved documentation

artem-smotrakov:
Part of #432


artem-smotrakov:
Hearing no objections.

PR #434, opened 2021-02-19
Improvements

artem-smotrakov:
This changeset proposes a bunch of minor updates such as fixed typos, improved output of the tool, renamed variables, etc


artem-smotrakov:
Hearing no objections.

Issue #436, opened 2021-02-23
PoC: Rating for open source rules of play


artem-smotrakov:
Implement a PoC for a rating that checks whether an open source project violates a set of rules (of play).

PR #438, opened 2021-02-23
Open source rules of play rating

artem-smotrakov:
Part of #436


src/main/java/com/sap/oss/phosphor/fosstars/tool/github/SingleSecurityRatingCalculator.java
```diff
@@ -5,144 +5,83 @@
...
 import java.util.Objects;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 /**
- * The class calculates a security rating for a single open-source project.
+ * The class calculates a rating for a project.
  */
-class SingleSecurityRatingCalculator {
+public class SingleSecurityRatingCalculator implements RatingCalculator {
```
mibo, line 21: Maybe I'm wrong, but should the class be renamed in e.g. `SingleRatingCalculator`?

```suggestion
public class SingleRatingCalculator implements RatingCalculator {
```

mibo (approved):
Only a minor comment.
Please check if valid otherwise all looks good (hence I approve it in advance 😉).

artem-smotrakov:
Thanks for the review @mibo !

PR #442, opened 2021-03-05
Reduce noise for maven download logging

mibo:


mibo:
Not sure why but the GH actions behave a bit different.
Hence I have set the log level for the `Slf4jMavenTransferListener` to `WARN` and set the CLI option `-ntp` (short for `--no-transfer-progress`).

artem-smotrakov (approved):
Looks good. We can do the same in https://github.com/SAP/fosstars-rating-core-action/blob/main/entrypoint.sh#L43

mibo:
> Looks good. We can do the same in https://github.com/SAP/fosstars-rating-core-action/blob/main/entrypoint.sh#L43

Good idea, will create a PR (later).

PR #467, opened 2021-03-23
SimpleAdvice class should be serializable

### LINKED ISSUE #380: SimpleAdvice class should be serializable
The class should support serialization to JSON/YAML.

artem-smotrakov:
Fixes #380


mibo (approved):
LGTM 👍

PR #496, opened 2021-04-08
Advice storage should not depend on a rating

### LINKED ISSUE #474: Improve how OssAdviceContentYamlStorage get information about all features
In the `com.sap.oss.phosphor.fosstars.advice.oss.OssAdviceContentYamlStorage` the storage is initialized based on all features which are used in a give rating:
```
static {
    try {
      DEFAULT = OssAdviceContentYamlStorage.loadFromResource(
          RESOURCE_PATH, RatingRepository.INSTANCE.rating(OssArtifactSecurityRating.class));
    } catch (IOException e) {
      throw new UncheckedIOException("Could not load advice", e);
    }
  }
```

See in addition: `com.sap.oss.phosphor.fosstars.advice.AdviceContentYamlStorage#loadFromResource`

This may lead to not available features if not all `com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures` are used in the given rating.

Hence, it would be better if it is initialized with all features defined in `OssFeatures`

artem-smotrakov:
Fixes #474


artem-smotrakov:
Hearing no objections.

PR #508, opened 2021-04-23
Introduce new VulnerableArtifact to hold CPE information

### LINKED ISSUE #487: Introduce VulnerableArtifact for Vulnerability
Introduce `VulnerableArtifact` for `Vulnerability` which holds the name (CPE?) and the version range for an artifact which is affected by they `Vulnerability`.
The `Vulnerability` would then hold a list of `VulnerableArtifact` (or maybe `AffectedArtifact`?).

sourabhsparkala:
- Introduce new class VulnerableArtifact
- Update and add new tests

Fixes #487

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/value/VersionRange.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/value/VersionRange.java
@@ -1,6 +1,7 @@
 package com.sap.oss.phosphor.fosstars.model.value;
 
-import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Objects;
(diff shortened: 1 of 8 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/model/value/VersionRange.java
```diff
@@ -10,42 +11,61 @@
...
-   * Both versions are including.
-   * Start and end version may be null.
+   * First vulnerable version (excluding).
+   */
+  private final String versionStartExcluding;
+
+  /**
+   * Last vulnerable version (excluding).
+   */
+  private final String versionEndExcluding;
```
artem-smotrakov, line 35: I am not sure if we need these fields (please note that this class doesn't model data from NVD -- we have classes in the `nvd` package that do that). Plus, the class doesn't check that all these four fields are correct (overlaps, etc). If you still would like to make this class aware of open/closed ends, I would suggest to use flags like in https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/oss/phosphor/fosstars/model/math/DoubleInterval.java

src/main/java/com/sap/oss/phosphor/fosstars/model/value/Vulnerability.java
```diff
@@ -104,9 +105,9 @@
   private final Date published;
 
   /**
-   * List of vulnerable version ranges.
+   * List of vulnerable artifacts.
    */
-  private final List<VersionRange> vulnerableVersions;
+  private final List<VulnerableArtifact> vulnerableArtifacts;
```
artem-smotrakov, line 110: This new class refers to NVD. But the `Vulnerability` class doesn't represent data from NVD. In fact, it describes a vulnerability, not an entry from NVD. NVD is not the only available database of vulnerabilities. Let's please not make `Vulnerability` specific to NVD.

I am not sure if `Vulnerability` needs to hold a list of artifacts (please see https://github.com/SAP/fosstars-rating-core/issues/487#issuecomment-809531836). This class already holds quite a lot of info. Plus, I am not sure if wee need `VulnerableArtifact` at all. You introduce `Artifact` and its implementations in another pull request. How about adding `AbstractArtifact` that implements `Artifact` and holds a list of vulnerabilities?

sourabhsparkala:
@mibo we decided to close this PR. Do you agree?

mibo:
> @mibo we decided to close this PR. Do you agree?

Based on the latest changes (e.g. in #534 and others) this PR and the related issue #487 are not relevant anymore.  
Hence, yes this can be closed.

PR #513, opened 2021-04-30
Make AbstractValue.isUnknown() final

### LINKED ISSUE #509: ArtifactVersionsValue.isUnknown() should return false
https://github.com/SAP/fosstars-rating-core/blob/e21dca466de6bbac96757d983d2010c2dad0b64e/src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersionsValue.java#L40

Currently, an `ArtifactVersionsValue` is known if its `ArtifactVersions` is not empty. However, an empty collection is a valid known value that just means that there is no artifact versions.

### LINKED ISSUE #512: Make AbstractValue.isUnknown() final
`AbstractValue` is supposed to be used for known values. `Feature.unknown()` or `UnknownValue` should be used to create unknown values.

artem-smotrakov:
Fixes #512
Fixes #509

### DIFF
--- a/src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReleasesFromGitHubTest.java
+++ b/src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReleasesFromGitHubTest.java
@@ -8,7 +8,9 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.sap.oss.phosphor.fosstars.model.Value;
 import com.sap.oss.phosphor.fosstars.model.subject.oss.GitHubProject;
+import com.sap.oss.phosphor.fosstars.model.value.ArtifactVersions;
 import com.sap.oss.phosphor.fosstars.model.value.ValueHashSet;
 import java.io.IOException;
 import java.util.Arrays;
@@ -118,6 +120,9 @@ public class ReleasesFromGitHubTest extends TestGitHubDataFetcherHolder {
     assertEquals(1, values.size());
     assertTrue(values.has(RELEASED_ARTIFACT_VERSIONS));
     assertTrue(values.of(RELEASED_ARTIFACT_VERSIONS).isPresent());
-    assertTrue(values.of(RELEASED_ARTIFACT_VERSIONS).get().isUnknown());
+
+    Value<ArtifactVersions> releasedArtifactVersions = values.of(RELEASED_ARTIFACT_VERSIONS).get();
+    assertFalse(releasedArtifactVersions.isUnknown());
+    assertTrue(releasedArtifactVersions.get().empty());
   }
 }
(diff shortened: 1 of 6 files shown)


src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReleasesFromGitHubTest.java
```diff
@@ -118,6 +120,9 @@ public void testIfGitHubHasNoReleaseAndTags() throws IOException {
     assertEquals(1, values.size());
     assertTrue(values.has(RELEASED_ARTIFACT_VERSIONS));
     assertTrue(values.of(RELEASED_ARTIFACT_VERSIONS).isPresent());
-    assertTrue(values.of(RELEASED_ARTIFACT_VERSIONS).get().isUnknown());
```
sourabhsparkala, line 121: I understand that we need the features to be implicitly a known value. But why was this line in test method removed? Why not in other tests like `LgtmDataProviderTest`. 

I am not sure I understand the idea behind this pull request.

sourabhsparkala (approved):
Just a query. If you feel that the changes are not required in other tests. Please merge, Everything else looks good 👍

src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReleasesFromGitHubTest.java
artem-smotrakov: I'll revert this line and replace `assertTrue` with `assertFalse`.

> I am not sure I understand the idea behind this pull request.

`AbstractValue` is supposed to be used as a base class for known values. Therefore, its `isUnknown()` method should return false and child classes should not be able to change that. It should have been called `AbstractKnownValue`.

artem-smotrakov: > I understand that we need the features to be implicitly a known value. But why was this line in test method removed? Why not in other tests like `LgtmDataProviderTest`.

This test case uses a project that has empty list of releases and tags on GitHub. Currently, it is treated as there is no info about releases for this project. But what if a project just has really no releases? In this case the feature should just be empty. It should not be unknown. What do you think?

sourabhsparkala: I agree, but 
- But sending an empty list of releases if releases are actually unknown or unfound also does not make sense.

I leave this up to you.

artem-smotrakov: > But sending an empty list of releases if releases are actually unknown or unfound also does not make sense. 

The difficult part here is that the provider can't actually distinguish between unknown release info and not found release info. I see you suggest to interpret an empty release info as unknown. I don't think it is good because:

- The current provider doesn't provide a correct value if a project just doesn't have release history
- Unknown values hurt confidentiality of scores. If case no release history actually exists, confidentiality of score would be lower than it actually is. That increases a chance to get `UNCLEAR` label when it is actually clear.

PR #515, opened 2021-05-03
OssArtifactSecurityRatingVerificationTest should work with dynamic dates

### LINKED ISSUE #484: Fix OssArtifactSecurityRatingVerificationTest to work with dynamic dates
Currently the `OssArtifactSecurityRatingVerificationTest` use the approach to defined the test vectors and expected outcome in a yaml file (`OssArtifactSecurityRatingTestVectors.yml`).
But this does not support the definition of dynamic dates (e.g. today - 90 days) which is required to reliably test e.g. the `ArtifactAgeScore` or `ArtifactReleaseHistoryScore`.

One solution is to migrate `OssArtifactSecurityRatingVerificationTest` to a programmatic test.
Another solution may be to support dynamic dates in the test vectors (without any changes in not test source code).

artem-smotrakov:

- Added `TestArtifactVersion` that takes an age of artifact and calculates its release date.
- Updated the test vectors for `OssArtifactSecurityRating` to use `TestArtifactVersion`.
- Added more javadoc, renamed several methods and variables.
- Added more test cases for serialization.

Fixes #484

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -52,8 +52,8 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
         ArtifactVersions.sortByReleaseDate(artifactVersions);
 
     // check release frequency over time
-    Collection<VersionInfo> versionInfos = createVersionInfos(sortedByReleaseDate);
-    VersionStats stats = calculateVersionStats(versionInfos);
+    Collection<VersionInfo> versionInfo = createVersionInfo(sortedByReleaseDate);
+    VersionStats stats = calculateVersionStats(versionInfo);
 
     if (stats.averageDaysBetweenReleases < 10) {
       scoreValue.increase(3);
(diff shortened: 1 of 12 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
```diff
@@ -113,30 +113,30 @@ static VersionStats calculateVersionStats(Collection<VersionInfo> versionInfos)
   /**
    * Create the version info based on list of sorted artifact versions.
    *
-   * @param artifactVersions sorted collection of artifact versions
-   * @return version infos
+   * @param artifactVersions A sorted collection of artifact versions.
+   * @return Version info.
```
sourabhsparkala, line 117: A collection of `VersionInfo`?

sourabhsparkala: Or did you mean to say that `VersionInfo` is itself a `List`?

src/test/resources/com/sap/oss/phosphor/fosstars/model/rating/oss/OssArtifactSecurityRatingTestVectors.yml
```diff
@@ -512,19 +517,18 @@ elements:
...
               introduced: "2019-01-01"
-              fixed: "2019-01-03"
               published: "2019-01-03"
               vulnerableVersions:
                 - versionStart: 1.0.0
                   versionEnd: 1.1.0
     expectedScore:
       type: "DoubleInterval"
-      from: 4.0
+      from: 0.0
```
sourabhsparkala, line 528: The expected score here had to be re-iterated because of `UNPATCHED` or something else? Let me know.

sourabhsparkala (approved):
Looks good 👍

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
artem-smotrakov: I mean information (info) about versions. If I understand correctly, we can't say "informations" or "infos"

https://dictionary.cambridge.org/grammar/british-grammar/information

sourabhsparkala: How about `VersionsInfo`? Just a way to indicate that it is a `Collection` of some sort.

src/test/resources/com/sap/oss/phosphor/fosstars/model/rating/oss/OssArtifactSecurityRatingTestVectors.yml
artem-smotrakov: The test vector failed because the rating procedure produces a score ~4.54 and therefore `MODERATE` label. But the test vector expects `BAD`. I suspect that we disabled this test before we updated the thresholds for labels when we introduced the rating. I made vulnerability unpatched because it reduces the security score. That was enough to pass the test vector.

@mibo Could you please have a look at it as you are the author of these test vectors?

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
artem-smotrakov: I updated it to "A collection of {@link VersionInfo}."

I don't think that is strictly necessary to tell what kind of class is returned by a method. I think it is more important to describe what kind of data it returns. In this case, the method returns information about artifacts. That's why it think "Version info" was enough.

src/test/resources/com/sap/oss/phosphor/fosstars/model/rating/oss/OssArtifactSecurityRatingTestVectors.yml
mibo: @artem-smotrakov Yes, we set the test to ignore before the final thresholds were defined.
At this time the BAD artifact rating had a higher impact to the combined rating.
The changes you did are fine as for this test a BAD project security rating combined with a GOOD artifact rating should result in a BAD combined rating  ("# Test vectors for: BAD project security, GOOD artifact security").

mibo (approved):

artem-smotrakov:
@sourabhsparkala @mibo Thanks for the review!

PR #516, opened 2021-05-04
Flexible serialization for implementations of main interfaces

### LINKED ISSUE #514: Flexible serialization for implementations of main interfaces
Currently, the main interfaces in the package `com.sap.oss.phosphor.fosstars.model` use `@JsonSubTypes` annotation to list all implementations for deserialization. Otherwise, Jackson would fail to deserialize them. It means that no other implementation of those interfaces can be deserialized. To overcome this, let's try to use  `ObjectMapper.registerSubtypes()`. The main implementations can be registered by `Json` and `Yamp` classes. Then, callers can register additional sybtypes if necessary. Unfortunately, `ObjectMapper.registerSubtypes()` doesn't work if an interface has `@JsonSubTypes`.

artem-smotrakov:
Currently, the main interfaces in the package `com.sap.oss.phosphor.fosstars.model` use `@JsonSubTypes` annotation to list all implementations for deserialization. Otherwise, Jackson would fail to deserialize them. It means that no other implementation of those interfaces can be deserialized.

To overcome this, `Json` and `Yaml` classes now use `ObjectMapper.registerSubtypes()` to register known subtypes for deserialization. Then, callers can register additional sybtypes if necessary.

- Get rid of `@JsonSubTypes`
- Updated `Json` and `Yaml` to return an `ObjectMapper` with registered default subtypes.
- Fixed a few tests that create their own `ObjectMapper`.

Fixes #514

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/Feature.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/Feature.java
@@ -1,27 +1,6 @@
 package com.sap.oss.phosphor.fosstars.model;
 
-import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import com.sap.oss.phosphor.fosstars.model.feature.BooleanFeature;
-import com.sap.oss.phosphor.fosstars.model.feature.BoundedDoubleFeature;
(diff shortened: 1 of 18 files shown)


sourabhsparkala (approved):
Looks good 👍

artem-smotrakov:
> Looks good

Thank you for the review @sourabhsparkala !

Issue #518, opened 2021-05-05
Subject support for Artifacts and updated CLI for multi-Subject handling


mibo:
Because of the changes in #504 the CLI need to be updated to support the new DataProvider which get release information for Artifacts.
During work on #505 I made me some thought and created (on top of the 505 PR/branch) some minor [PoC in my repo](https://github.com/mibo/fosstars-rating-core/pull/10).
It is farm from the shape to be merged (very early draft), but I thought it may be interesting to discuss if this is a valid option to add subject support for Artifacts and updated the CLI for multi-Subject handling.

artem-smotrakov:
I also think that an artifact should be `Subject`. In your PoC, you also added support for NPM packages in the command-line tool. That's also good.

mibo:
Thanks for the feedback.
After #505 is closed I will rebase my PoC on the master and create a PR.
Then we can see how we continue with the PoC 🙂

PR #522, opened 2021-05-05
Subject support to Artifacts and updated CLI for multi-Subject handling

mibo:
**POC**: Subject support to Artifacts and updated CLI for multi-Subject handling.

This PR is related to the issue #518 and contains the mentioned PoC.
It is in an **early PoC state used to get feedback**.


src/main/java/com/sap/oss/phosphor/fosstars/data/DataProvider.java
```diff
@@ -65,6 +65,8 @@
    */
   Set<Feature<?>> supportedFeatures();
 
+  boolean supports(Object type);
```
artem-smotrakov, line 68:
```suggestion
  boolean supports(T type);
```

artem-smotrakov, line 68: Let's please try to avoid using `Object`.

```suggestion
  boolean supports(T type);
```

Or, maybe it is even better if we get rid of `T` and use `Subject` instead of it:

```java
public interface DataProvider {
  /**
   * Gathers data about an object and updates a specified {@link ValueSet}.
   *
   * @param object The object.
   * @param values The set of values to be updated.
   * @return This data provider.
   * @throws IOException If something went wrong.
   */
  DataProvider update(Subject subject, ValueSet values) throws IOException;
...

  boolean supports(Subject subject);
...
```

mibo: > Let's please try to avoid using `Object`.

Completely agree and I like the idea using the `Subject`.
Have updated the PR/PoC accordingly.

artem-smotrakov (approved):
Let's merge it to a separate branch and continue working on it.

PR #535, opened 2021-05-11
Use GitHub License API in LicenseInfo data provider

### LINKED ISSUE #525: Use GitHub License API in LicenseInfo data provider
Currently, the data provider `LicenseInfo` uses a list of known locations when it searches for a license file, and then it uses a regex to check whether the license is allowed or not (the `ALLOWED_LICENSE` feature). It might be good to add an option to use [GitHub License API](https://docs.github.com/en/rest/reference/licenses) for that.

artem-smotrakov:
GitHub API may be a preferable source of info for the OSS RoP rating.

Fixes #525

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/data/github/LicenseInfo.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/data/github/LicenseInfo.java
@@ -8,6 +8,7 @@ import static com.sap.oss.phosphor.fosstars.util.Deserialization.readListFrom;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.sap.oss.phosphor.fosstars.model.Feature;
+import com.sap.oss.phosphor.fosstars.model.Value;
 import com.sap.oss.phosphor.fosstars.model.ValueSet;
 import com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures;
 import com.sap.oss.phosphor.fosstars.model.subject.oss.GitHubProject;
@@ -25,6 +26,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import org.kohsuke.github.GHLicense;
+import org.kohsuke.github.GitHub;
+import org.kohsuke.github.GitHubBuilder;
 
 /**
  * This data provider gathers info about project's license. It fills out the following features:
@@ -35,6 +39,13 @@ import java.util.stream.Collectors;
  * </ul>
  */
 public class LicenseInfo extends GitHubCachingDataProvider {
+
+  /**
+   * Possible sources of information for the data provider.
+   */
+  enum Source {
+    LOCAL_REPOSITORY, GITHUB_API
+  }
 
   /**
    * A list of files that may be a license.
@@ -56,6 +67,16 @@ public class LicenseInfo extends GitHubCachingDataProvider {
    * A list of patterns that are not allowed in licenses.
    */
   private final List<Pattern> disallowedLicensePatterns = new ArrayList<>();
+
+  /**
+   * A list of allowed license names that are known on GitHub.
+   */
+  private final List<String> githubNamesOfAllowedLicenses = new ArrayList<>();
+
+  /**
+   * Preferred source of information for the data provider.
+   */
+  private Source preferredSource = Source.LOCAL_REPOSITORY;
 
   /**
    * Initializes a data provider. The constructor searches for default configs for the provider.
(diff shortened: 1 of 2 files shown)


SebastianWolf-SAP (commented):
Overall approve the direction where this change is going. However, have some concerns about some implementation details:
- Do we really need the source of information or can we make the GitHub API default?
- According to my local tests (might be wrong), the library for the GitHub API doesn't really retrieve the content of the license file from the repository, but the standard license from a remote server which makes it impossible to detect if it contains invalid content. That would make the disallowed content check pointless if we use the GitHub API library
- If we are able to identify the licenses by an ID (either the GitHub ID or SPDX ID), we probably don't need patterns anymore and can use a list of IDs in the configuration to make it work properly

Actually, I have a parallel implementation to this issue in the works (I thought it might be a nice one to get myself familiar with FOSSTARS ;)) that already includes the points I mentioned above, but is not tested yet (except the unit tests of course). See my fork and the change https://github.com/SebastianWolf-SAP/fosstars-rating-core/commit/5699b558d2c982081c7c862c79fffb240d5cb0b2 for details. No idea when I will have time to complete that with testing etc. but maybe you want to have a look already and include some things into this change.

artem-smotrakov:
Thanks for the review @SebastianWolf-SAP !

> * Do we really need the source of information or can we make the GitHub API default?

We currently have to main ways for gathering info about projects on GitHub:

1. Using GitHub API
2. Cloning the project's repository and working with the local files

Although the first way often allows to gather all necessary info, this way has one big disadvantage -- it requires a GitHub token and tokens have rate/connection limits. While gathering data about multiple projects, it is quite easy to hit the limits. Therefore, many data providers just clone repositories to save the token's capabilities for the providers that require using GitHub API. I'd prefer `LicenceInfo` to implement both ways in case we'd like to use this data provider later for collecting data for other rating.

The default source of info for `LicenceInfo` was set to `Source.LOCAL_REPOSITORY` because I thought it might be better to make the provider frugal by default :) However, I don't really mind changing the default mode to the GitHub API.

> * According to my local tests (might be wrong), the library for the GitHub API doesn't really retrieve the content of the license file from the repository, but the standard license from a remote server which makes it impossible to detect if it contains invalid content. That would make the disallowed content check pointless if we use the GitHub API library

That's a very good point. Let me check it.

> * If we are able to identify the licenses by an ID (either the GitHub ID or SPDX ID), we probably don't need patterns anymore and can use a list of IDs in the configuration to make it work properly

Yeah, I think we can add such an option.

> Actually, I have a parallel implementation to this issue in the works (I thought it might be a nice one to get myself familiar with FOSSTARS ;)) that already includes the points I mentioned above, but is not tested yet (except the unit tests of course). See my fork and the change [SebastianWolf-SAP@5699b55](https://github.com/SebastianWolf-SAP/fosstars-rating-core/commit/5699b558d2c982081c7c862c79fffb240d5cb0b2) for details. No idea when I will have time to complete that with testing etc. but maybe you want to have a look already and include some things into this change.

Great! Let me have a look. Sorry, I didn't know that you work on it. If I knew, I would love to let you finish it :)

artem-smotrakov:
> > Actually, I have a parallel implementation to this issue in the works (I thought it might be a nice one to get myself familiar with FOSSTARS ;)) that already includes the points I mentioned above, but is not tested yet (except the unit tests of course). See my fork and the change [SebastianWolf-SAP@5699b55](https://github.com/SebastianWolf-SAP/fosstars-rating-core/commit/5699b558d2c982081c7c862c79fffb240d5cb0b2) for details. No idea when I will have time to complete that with testing etc. but maybe you want to have a look already and include some things into this change.
> 
> Great! Let me have a look. Sorry, I didn't know that you work on it. If I knew, I would love to let you finish it :)

@SebastianWolf-SAP I had a look at your version of `LicenseInfo` and I like it more than mine :) I think it would be better if we integrated your updates. Maybe you can already open a pull request -- I might be missing something but the code looks pretty good to me. Maybe we can fix some bugs later if they escape.

SebastianWolf-SAP:
Well, I wasn't talking about it as I finally wanted to get started and didn't know how fast I could have something in my hands (due to the fact that I'm constantly in meetings) or if it would work at all. But next time, we should definitely align as it seems to work. :) 

Thanks for the feedback! I'd be glad to have my contribution integrated upstream, but I also understand your concerns regarding the API limitation. Could we potentially automate that, i.e. automatically switching to the repository mode if no token is provided or if a test call returns a 403? Background is that I'd like to have at least the license check in a standardized way if possible as otherwise we will receive complaints by our users if the results are not matching the info which is displayed on GitHub itself. 

I'll try to make the changes in my branch as soon as possible, but it seems that I'll be pretty busy with other stuff until Wednesday, but let's see, maybe I can squeeze it in. First, I also need to make my IDE work according to your checkstyle rules...

artem-smotrakov:
> Thanks for the feedback! I'd be glad to have my contribution integrated upstream, but I also understand your concerns regarding the API limitation. Could we potentially automate that, i.e. automatically switching to the repository mode if no token is provided or if a test call returns a 403? Background is that I'd like to have at least the license check in a standardized way if possible as otherwise we will receive complaints by our users if the results are not matching the info which is displayed on GitHub itself.

Maybe we won't even hit the limits, so I think we can implement the fallback to local repositories later if necessary.

> I'll try to make the changes in my branch as soon as possible, but it seems that I'll be pretty busy with other stuff until Wednesday, but let's see, maybe I can squeeze it in. First, I also need to make my IDE work according to your checkstyle rules...

Sure, no problem. Let me know if you need help.

artem-smotrakov:
Fixed in https://github.com/SAP/fosstars-rating-core/pull/550

PR #538, opened 2021-05-12
Check if README is incomplete in the OSS RoP rating

### LINKED ISSUE #526: Check if README is incomplete in the OSS RoP rating
Things to do:
- [x] Add a new feature that tells that README is incomplete i.e., it violates certain rules.
- [x] Include the new feature to the OSS RoP rating.
- [x] Implement a data provider for the new feature. The provider should check it README follows certain rules. The rules should be configurable.
- [x] The new feature should have `rl-readme_file-2` identifier in the report.

### LINKED ISSUE #537: Release unreleased file streams
There are multiple I/O file streams that are not closed. Look for `newInputStream()` and `LocalRepository.read()` calls. I/O file streams should be closed.

artem-smotrakov:

- Added a new feature `INCOMPLETE_README`.
- Included the new feature to the OSS RoP score.
- Updated `ReadmeInfo` data provider to support the new feature.
- Updated tests.
- Released unreleased file streams, this fixes #537 

Closes #526

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/data/github/CodeqlDataProvider.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/data/github/CodeqlDataProvider.java
@@ -78,13 +78,14 @@ public class CodeqlDataProvider extends GitHubCachingDataProvider {
     // ideally, we're looking for a GitHub action that runs CodeQL scan on pull requests
     // but if we just find an action that runs CodeQL scans, that's also fine
     for (Path configPath : findGitHubActionsIn(repository)) {
-      InputStream content = Files.newInputStream(configPath);
-      Map<String, Object> githubAction = Yaml.readMap(content);
-      if (triggersCodeqlScan(githubAction)) {
-        runsCodeqlScans = RUNS_CODEQL_SCANS.value(true);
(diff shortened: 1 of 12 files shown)


SebastianWolf-SAP (approved):
Looks good to me, thanks for the extension. One thing maybe: With this setup it's not really transparent to the user which sections in the README are missing, so that might probably be a thing we could/should add as well. Moreover, the set of patterns probably needs to be aligned with the ones used in the Repository Linter.

But both can also be done later (and I would also volunteer to do that). ;)

artem-smotrakov:
Thanks for the review @SebastianWolf-SAP !

> With this setup it's not really transparent to the user which sections in the README are missing, so that might probably be a thing we could/should add as well.

Good point! I just created #539 to address it.

> Moreover, the set of patterns probably needs to be aligned with the ones used in the Repository Linter.
> But both can also be done later 

Yes, I think the patterns need to be improved.

> (and I would also volunteer to do that). ;)

That would be great :) Looking forward to seeing your pull requests! You can find the current configs in the branch https://github.com/SAP/fosstars-rating-core/tree/fosstars-oss-rules-of-play-sap-report

PR #542, opened 2021-05-17
Better Markdown formatters

### LINKED ISSUE #501: Show advice higher in Markdown reports
Currently, the existing Markdown formatters print advice on the bottom of the report. As a result, it is likely that the users don't find the advice. It would be better if the advice were somewhere higher in the report.

artem-smotrakov:
Here is a list of updates:

- Updated formatters to accept custom templates.
- Show advice higher in Markdown reports for `OssSecurityRating`.
- Updated `CommonFormatter` to use `Precision.equals()` when comparing doubles.

Fixes #501


sourabhsparkala (approved):
👍

Issue #480, opened 2021-03-25
Improve ArtifactVersionVulnerabilityScore to work with not semantic versions


mibo:
Improve `ArtifactVersionVulnerabilityScore` to work with not semantic versions.
Currently only version which follows the [Semantic Version](https://semver.org/) format can be used.

PR #549, opened 2021-05-18
Add technique to compare semantic versions

### LINKED ISSUE #479: Improve ArtifactReleaseHistoryScore by considering semantic versions
Improve `ArtifactReleaseHistoryScore` by considering semantic versions.
Currently there is no distinction in the release history between major, minor or micro releases.
Only the release date of all known releases are used.

This can be improved by calculation of a release history score for each major version.
All of those builds the final score together with the score over all releases.

A further improvement is to consider for which artifact major version the rating is calculated.

The new functionality must be covered with according test cases.

sourabhsparkala:
- Implement filtering technique for artifact versions by major version.
- Update and add tests.

Fixes #479

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -166,7 +166,7 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
     if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
       ArtifactVersions filteredArtifactVersions =
           artifactVersions.get().filterArtifactsByMajorVersion(
-              artifactVersion.get().getSemanticVersion().get().toString());
+              artifactVersion.get().getSemanticVersion().get());
 
       if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
         return filteredArtifactVersions.sortByReleaseDate();
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -57,13 +57,7 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
     ScoreValue scoreValue = scoreValue(6.0, artifactVersions, artifactVersion);
 
     final Collection<ArtifactVersion> artifactCollection =
-        filter(artifactVersions, artifactVersion);
-
-    // If the artifact collection is filtered, then a cluster of versions have been obtained.
-    // Increase the score value.
-    if (artifactCollection.size() < artifactVersions.get().size()) {
-      scoreValue.increase(1);
-    }
+        filter(artifactVersions, artifactVersion, scoreValue);
 
     // check release frequency over time
     Collection<VersionInfo> versionInfo = versionInfo(artifactCollection);
@@ -159,16 +153,18 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
    * 
    * @param artifactVersions {@link ArtifactVersions} to be filtered.
    * @param artifactVersion {@link ArtifactVersion} provided by the user.
+   * @param scoreValue {@link ScoreValue} to be updated on finding a cluster of versions.
    * @return A collection of {@link ArtifactVersion}.
    */
   private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
-      Value<ArtifactVersion> artifactVersion) {
+      Value<ArtifactVersion> artifactVersion, ScoreValue scoreValue) {
     if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
-      ArtifactVersions filteredArtifactVersions =
-          artifactVersions.get().filterArtifactsByMajorVersion(
-              artifactVersion.get().getSemanticVersion().get());
+      ArtifactVersions filteredArtifactVersions = artifactVersions.get()
+          .filterArtifactsByMajorVersion(artifactVersion.get().getSemanticVersion().get());
 
+      // If a cluster of versions has been obtained. Increase the score value.
       if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
+        scoreValue.increase(1);
         return filteredArtifactVersions.sortByReleaseDate();
       }
     }
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -26,6 +26,12 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
    * The artifact versions list should have size greater than the threshold.
    */
   private static final int ARTIFACT_VERSIONS_SIZE_THRESHOLD = 1;
+  
+  /**
+   * The artifact versions list to be considered as a cluster should have size greater than the
+   * threshold.
+   */
+  private static final int VERSIONS_CLUSTER_SIZE_THRESHOLD = 2;
 
   /**
    * Initializes a new score.
@@ -54,10 +60,10 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
           .withMinConfidence();
     }
 
-    ScoreValue scoreValue = scoreValue(6.0, artifactVersions, artifactVersion);
+    ScoreValue scoreValue = scoreValue(7.0, artifactVersions, artifactVersion);
 
     final Collection<ArtifactVersion> artifactCollection =
-        filter(artifactVersions, artifactVersion, scoreValue);
+        filter(artifactVersions, artifactVersion);
 
     // check release frequency over time
     Collection<VersionInfo> versionInfo = versionInfo(artifactCollection);
@@ -149,22 +155,21 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
 
   /**
    * Filter the artifact versions matching with the major version of the given
-   * {@link ArtifactVersion}. Return the filtered list if size > 1.
+   * {@link ArtifactVersion}. Return the filtered list if size >
+   * {@link #VERSIONS_CLUSTER_SIZE_THRESHOLD}.
    * 
    * @param artifactVersions {@link ArtifactVersions} to be filtered.
    * @param artifactVersion {@link ArtifactVersion} provided by the user.
-   * @param scoreValue {@link ScoreValue} to be updated on finding a cluster of versions.
    * @return A collection of {@link ArtifactVersion}.
    */
   private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
-      Value<ArtifactVersion> artifactVersion, ScoreValue scoreValue) {
+      Value<ArtifactVersion> artifactVersion) {
     if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
       ArtifactVersions filteredArtifactVersions = artifactVersions.get()
           .filterArtifactsByMajorVersion(artifactVersion.get().getSemanticVersion().get());
 
-      // If a cluster of versions has been obtained. Increase the score value.
-      if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
-        scoreValue.increase(1);
+      // If a cluster of versions has been obtained, return the clustered artifact versions.
+      if (filteredArtifactVersions.size() > VERSIONS_CLUSTER_SIZE_THRESHOLD) {
         return filteredArtifactVersions.sortByReleaseDate();
       }
     }
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -1,5 +1,6 @@
 package com.sap.oss.phosphor.fosstars.model.score.oss;
 
+import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.ARTIFACT_VERSION;
 import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.RELEASED_ARTIFACT_VERSIONS;
 import static java.time.temporal.ChronoUnit.DAYS;
 
@@ -22,16 +23,23 @@ import java.util.Iterator;
 public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
 
   /**
+   * The artifact versions list should have size greater than the threshold.
+   */
+  private static final int ARTIFACT_VERSIONS_SIZE_THRESHOLD = 1;
+
+  /**
    * Initializes a new score.
    */
   public ArtifactReleaseHistoryScore() {
     super("How frequent an open source project releases new versions",
-        OssFeatures.RELEASED_ARTIFACT_VERSIONS);
+        RELEASED_ARTIFACT_VERSIONS,
+        ARTIFACT_VERSION);
   }
 
   @Override
   public ScoreValue calculate(Value<?>... values) {
     Value<ArtifactVersions> artifactVersions = find(RELEASED_ARTIFACT_VERSIONS, values);
+    Value<ArtifactVersion> artifactVersion = find(ARTIFACT_VERSION, values);
 
     if (artifactVersions.isUnknown()) {
       return scoreValue(0.0, artifactVersions)
@@ -40,19 +48,19 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
           .explain("No versions are found. Hence, no release history score can be calculated");
     }
 
-    if (artifactVersions.get().size() <= 1) {
+    if (artifactVersions.get().size() <= ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
       return scoreValue(0.0, artifactVersions)
           .explain("Only one version is given. Hence, no release history score can be calculated")
           .withMinConfidence();
     }
 
-    ScoreValue scoreValue = scoreValue(7.0, artifactVersions);
+    ScoreValue scoreValue = scoreValue(6.0, artifactVersions, artifactVersion);
 
-    Collection<ArtifactVersion> sortedByReleaseDate =
-        ArtifactVersions.sortByReleaseDate(artifactVersions);
+    final Collection<ArtifactVersion> artifactCollection =
+        filter(artifactVersions, artifactVersion, scoreValue);
 
     // check release frequency over time
-    Collection<VersionInfo> versionInfo = versionInfo(sortedByReleaseDate);
+    Collection<VersionInfo> versionInfo = versionInfo(artifactCollection);
     VersionStats stats = calculateVersionStats(versionInfo);
 
     if (stats.averageDaysBetweenReleases < 10) {
@@ -137,6 +145,30 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
     }
 
     return versionInfo;
+  }
+
+  /**
+   * Filter the artifact versions matching with the major version of the given
+   * {@link ArtifactVersion}. Return the filtered list if size > 1.
+   * 
+   * @param artifactVersions {@link ArtifactVersions} to be filtered.
+   * @param artifactVersion {@link ArtifactVersion} provided by the user.
+   * @param scoreValue Increment the score if a filtered list is found.
+   * @return A collection of {@link ArtifactVersion}.
+   */
+  private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
+      Value<ArtifactVersion> artifactVersion, ScoreValue scoreValue) {
+    if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
+      ArtifactVersions filteredArtifactVersions =
+          artifactVersions.get().filterArtifactsByMajorVersion(
+              artifactVersion.get().getSemanticVersion().get().toString());
+
+      if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
+        scoreValue.increase(1);
+        return filteredArtifactVersions.sortByReleaseDate();
+      }
+    }
+    return artifactVersions.get().sortByReleaseDate();
   }
 
   /**
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
@@ -57,7 +57,13 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
     ScoreValue scoreValue = scoreValue(6.0, artifactVersions, artifactVersion);
 
     final Collection<ArtifactVersion> artifactCollection =
-        filter(artifactVersions, artifactVersion, scoreValue);
+        filter(artifactVersions, artifactVersion);
+
+    // If the artifact collection is filtered, then a cluster of versions have been obtained.
+    // Increase the score value.
+    if (artifactCollection.size() < artifactVersions.get().size()) {
+      scoreValue.increase(1);
+    }
 
     // check release frequency over time
     Collection<VersionInfo> versionInfo = versionInfo(artifactCollection);
@@ -153,18 +159,16 @@ public class ArtifactReleaseHistoryScore extends FeatureBasedScore {
    * 
    * @param artifactVersions {@link ArtifactVersions} to be filtered.
    * @param artifactVersion {@link ArtifactVersion} provided by the user.
-   * @param scoreValue Increment the score if a filtered list is found.
    * @return A collection of {@link ArtifactVersion}.
    */
   private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
-      Value<ArtifactVersion> artifactVersion, ScoreValue scoreValue) {
+      Value<ArtifactVersion> artifactVersion) {
     if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
       ArtifactVersions filteredArtifactVersions =
           artifactVersions.get().filterArtifactsByMajorVersion(
               artifactVersion.get().getSemanticVersion().get().toString());
 
       if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
-        scoreValue.increase(1);
         return filteredArtifactVersions.sortByReleaseDate();
       }
     }
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
@@ -15,6 +15,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
+import org.apache.maven.artifact.versioning.ComparableVersion;
 
 /**
  * A set versions for an artifact.
@@ -125,6 +126,54 @@ public class ArtifactVersions implements Iterable<ArtifactVersion> {
       return Collections.emptyList();
     }
     return artifactVersions.get().sortByReleaseDate();
+  }
+
+  /**
+   * Filter {@link ArtifactVersions} by major version.
+   *
+   * @param version The artifact version.
+   * @return Filtered {@link ArtifactVersions}.
+   */
+  public ArtifactVersions filterArtifactsByMajorVersion(String version) {
+    Optional<SemanticVersion> currentVersion = SemanticVersion.parse(version);
+    if (!currentVersion.isPresent()) {
+      return ArtifactVersions.ofNothing();
+    }
+
+    String currentMajorVersion = buildMajorVersion(currentVersion.get().getMajor());
+    Optional<String> nextMajorVersion = nextMajorVersionOf(version);
+
+    return new ArtifactVersions(elements.stream().filter(v -> v.hasValidSemanticVersion())
+        .filter(v -> new ComparableVersion(v.version())
+            .compareTo(new ComparableVersion(currentMajorVersion)) >= 0
+            && new ComparableVersion(v.version())
+                .compareTo(new ComparableVersion(nextMajorVersion.get())) < 0)
+        .collect(Collectors.toSet()));
+  }
+
+  /**
+   * Determine next major semantic version from the given version.
+   * 
+   * @param version from which the next major version should be determined.
+   * @return the determined next possible major version.
+   */
+  private static Optional<String> nextMajorVersionOf(String version) {
+    Optional<SemanticVersion> semanticVersion = SemanticVersion.parse(version);
+    if (!semanticVersion.isPresent()) {
+      return Optional.empty();
+    }
+
+    return Optional.of(buildMajorVersion(semanticVersion.get().getMajor() + 1));
+  }
+
+  /**
+   * Build major version release semantic format.
+   * 
+   * @param major Number to build the semantic version.
+   * @return Semantic version format.
+   */
+  private static String buildMajorVersion(int major) {
+    return String.format("%s.0.0", major);
   }
 
   /**
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
@@ -112,7 +112,7 @@ public class ArtifactVersions implements Iterable<ArtifactVersion> {
    *
    * @param versions The other versions.
    * @return True if all the other versions is present in the collection,
-   *          false otherwise.
+   *         false otherwise.
    */
   public boolean containsAllOf(ArtifactVersions versions) {
     for (ArtifactVersion version : versions) {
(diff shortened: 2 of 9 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
```diff
@@ -40,19 +48,19 @@ public ScoreValue calculate(Value<?>... values) {
...
 
-    if (artifactVersions.get().size() <= 1) {
+    if (artifactVersions.get().size() <= ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
       return scoreValue(0.0, artifactVersions)
           .explain("Only one version is given. Hence, no release history score can be calculated")
           .withMinConfidence();
     }
 
-    ScoreValue scoreValue = scoreValue(7.0, artifactVersions);
+    ScoreValue scoreValue = scoreValue(6.0, artifactVersions, artifactVersion);
```
artem-smotrakov, line 57: I am not sure I understand why the score should be lower if a project doesn't use semantic versions. Is it semantic versioning any better than other schemes?

I am not sure either why the score should be lower when release history for a major version is short. It can happen that a project has a long release history but it just recently shipped a new major version. In this case, I'd fallback to using the entire release history.

@mibo What do you think?

sourabhsparkala: As per my understanding. If we are calculating the score on more related Artifacts as compared to all the artifacts. This would avoid variations and would handle diversity better. I believe the more focused related list of artifacts should have a better score as compared to considering all the artifacts (which may or may not be related). The idea is to have an appropriate clustering before we calculate the score.

mibo: > I am not sure either why the score should be lower when release history for a major version is short. It can happen that a project has a long release history but it just recently shipped a new major version. In this case, I'd fallback to using the entire release history.

Afaik the current implementation is that in case of only one version for a major version exists (e.g. only `2.0.0`) it fallback to use the entires release history.

However, maybe the score can calculated the release trend for all versions and for the according major version and build a final trend out of both (just as an idea).

> I am not sure I understand why the score should be lower if a project doesn't use semantic versions. Is it semantic versioning any better than other schemes?

Idea was that if we have the semantic version that only the versions with the same major version are considered (the clustering mentioned by Sourabh).
As sample, someone want the score for version `1.2.0` (release one year after `1.1.0`) but the latest is `3.1.0` (release one month before 3.0.0) the score should be lower.
If the version is not a semantic versions this may lead to higher scores why we thought to start with a lower score value (`6`).

sourabhsparkala: I agree with @mibo. This way the focus would be more on the possible affected list of versions.

src/main/java/com/sap/oss/phosphor/fosstars/model/value/SemanticVersion.java
```diff
@@ -4,6 +4,7 @@
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.codehaus.plexus.util.StringUtils;
```
mibo, line 7: Was it intended to use `SpringUtils` from the Plexus/Codehaus library?
I would recommend to not add a dependency to such a library because of an `StringUtils.isEmpty(..)` check.

sourabhsparkala: It was intended. The version check was just for null. I wanted to check for the `""` empty string as well. If you feel that we do not need this check. And let the parse matcher take over. I do not mind that also. I will revert it back to the original.

sourabhsparkala: I did not add any new dependency though. It was part of something else.

sourabhsparkala: But now that I think about this. We do not need an empty string check. The parse matcher already does that for us.

mibo: Reason why I asked was that I recommend to only use classes of direct dependencies and not from indirect dependencies 😉
And to not add a new dependency only because of something like a null/empty check for a string.

Hope this way my comment is more clear 🙂
However, removing the check when it is not required is the best solution 😉

```
+- org.apache.maven:maven-model:jar:3.8.1:compile
|  \- org.codehaus.plexus:plexus-utils:jar:3.2.1:compile
```

mibo (approved):
For me it looks good.
But please wait with merging for approval by @artem-smotrakov

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
artem-smotrakov: > Afaik the current implementation is that in case of only one version for a major version exists (e.g. only 2.0.0) it fallback to use the entires release history.

@sourabhsparkala Do we have a test case for that?

> Idea was that if we have the semantic version that only the versions with the same major version are considered (the clustering mentioned by Sourabh).

@sourabhsparkala Do we have a test case for that? Do we have test cases with not-semantic versions?

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -98,8 +100,8 @@ public void testCurrentReleasedVersionBestHistory() {
     ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
     Value<ArtifactVersions> versions = RELEASED_ARTIFACT_VERSIONS.value(
         ArtifactVersions.of(version100, version110, version120));
-    ScoreValue value = score.calculate(versions);
-    Assert.assertEquals(9.5, value.get(), DELTA);
+    ScoreValue value = score.calculate(versions, ARTIFACT_VERSION.unknown());
```
artem-smotrakov, line 103: Why don't we use unknown artifact versions? I think it would be better to use real versions and add a few new test cases with unknown values.

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +241,33 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(31));
+    ArtifactVersion version210 = new ArtifactVersion("2.1.0", now);
+
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    Value<ArtifactVersions> versions =
+        RELEASED_ARTIFACT_VERSIONS.value(ArtifactVersions.of(version100, version110, version120,
+            version130, version140, version150, version200, version210));
+    Value<ArtifactVersion> version = ARTIFACT_VERSION.value(version120);
+    ScoreValue value = score.calculate(versions, version);
+    Assert.assertEquals(8.8, value.get(), DELTA);
```
artem-smotrakov, line 271: Could you please add `assertEquals()` for "2.0.0", "1.5.0" and "1.0.0"?

artem-smotrakov (changes requested):

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +128,54 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+   * @return Filtered {@link ArtifactVersions}.
+   */
+  public ArtifactVersions filterArtifactsByMajorVersion(String version) {
+    Optional<SemanticVersion> currentVersion = SemanticVersion.parse(version);
+    if (!currentVersion.isPresent()) {
+      return ArtifactVersions.ofNothing();
+    }
+
+    String currentMajorVersion = buildMajorVersion(currentVersion.get().getMajor());
+    Optional<String> nextMajorVersion = nextMajorVersionOf(version);
```
artem-smotrakov, line 144: Here you already know that `version` is semantic, so you can guess major branches without parsing `version` again in `nextMajorVersionOf()`.

```suggestion
    int major = currentVersion.get().getMajor();
    ComparableVersion currentMajorVersion = new ComparableVersion(String.format("%s.0.0", major));
    ComparableVersion nextMajorVersion = new ComparableVersion(String.format("%s.0.0", major + 1));
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +128,54 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+  /**
+   * Build major version release semantic format.
+   * 
+   * @param major Number to build the semantic version.
+   * @return Semantic version format.
+   */
+  private static String buildMajorVersion(int major) {
+    return String.format("%s.0.0", major);
+  }
+
```
artem-smotrakov, line 178: Please also see the suggestions above.

```suggestion
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +128,54 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+    }
+
+    String currentMajorVersion = buildMajorVersion(currentVersion.get().getMajor());
+    Optional<String> nextMajorVersion = nextMajorVersionOf(version);
+
+    return new ArtifactVersions(elements.stream().filter(v -> v.hasValidSemanticVersion())
+        .filter(v -> new ComparableVersion(v.version())
+            .compareTo(new ComparableVersion(currentMajorVersion)) >= 0
+            && new ComparableVersion(v.version())
+                .compareTo(new ComparableVersion(nextMajorVersion.get())) < 0)
```
artem-smotrakov, line 150: Could it be simplified a bit?

```suggestion
    return new ArtifactVersions(
        elements.stream()
            .filter(v::hasValidSemanticVersion)
            .map(v -> new ComparableVersion(v.version()))
            .filter(v -> v.compareTo(currentMajorVersion) >= 0 && v.compareTo(nextMajorVersion) < 0)
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/SemanticVersion.java
```diff
@@ -158,4 +157,14 @@ public boolean equals(Object o) {
...
     return Objects.hash(major, minor, micro);
   }
+
+  @Override
+  public String toString() {
+    if (micro > -1) {
+      return String.format("%s.%s.%s", major, minor, micro);
+    } else if (minor > -1) {
+      return String.format("%s.%s", major, minor);
+    }
```
artem-smotrakov, line 167: The values can't be negative, please see the constructor.

src/main/java/com/sap/oss/phosphor/fosstars/model/value/SemanticVersion.java
artem-smotrakov: Yes, please use `StringUtils` from Apache Commons.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
sourabhsparkala: We do not have any test cases for non-sematic. As `SemanticVersion` would fail. We have not figured a way to handle non-sematic as of now. This issue is #480

artem-smotrakov: > We do not have any test cases for non-sematic. As `SemanticVersion` would fail. We have not figured a way to handle non-sematic as of now. This issue is #480

The current scoring should not fail if it sees a non-semantic version. #480 is about improving the scoring function to handle not-semantic versions in a better way. I think it would be good to add tests for the statements from @mibo above. We'll see then what kind of scores the function produces.

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +128,54 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+  public ArtifactVersions filterArtifactsByMajorVersion(String version) {
+    Optional<SemanticVersion> currentVersion = SemanticVersion.parse(version);
+    if (!currentVersion.isPresent()) {
+      return ArtifactVersions.ofNothing();
+    }
+
+    String currentMajorVersion = buildMajorVersion(currentVersion.get().getMajor());
+    Optional<String> nextMajorVersion = nextMajorVersionOf(version);
+
+    return new ArtifactVersions(elements.stream().filter(v -> v.hasValidSemanticVersion())
```
sourabhsparkala, line 146: I have a concern with this check
- This will ignore the `ArtifactVersion`s which are non-semantic. Then return the collection for further processing.
- But as a fallback if the returned filtered list contains only single major version `2.0.0`, then that list will contain a non-semantic version.

Please let me know if this is alright? @artem-smotrakov @mibo

mibo: > But as a fallback if the returned filtered list contains only single major version 2.0.0, then that list will contain a non-semantic version.

Not sure if I understand this sentence.
If the list only contains the `2.0.0` (as the only major version and given as `version` parameter) then the filtered result will contain only this version which is a semantic version.
Why the list should then contain a non-semantic version?

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
```diff
@@ -139,6 +153,28 @@ static VersionStats calculateVersionStats(Collection<VersionInfo> versionInfoCol
...
+   * @param artifactVersions {@link ArtifactVersions} to be filtered.
+   * @param artifactVersion {@link ArtifactVersion} provided by the user.
+   * @return A collection of {@link ArtifactVersion}.
+   */
+  private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
+      Value<ArtifactVersion> artifactVersion) {
+    if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
+      ArtifactVersions filteredArtifactVersions =
+          artifactVersions.get().filterArtifactsByMajorVersion(
+              artifactVersion.get().getSemanticVersion().get().toString());
```
artem-smotrakov, line 169: `filterArtifactsByMajorVersion()` takes a string and then parses it to a semantic version. Instead, it can accept `SemanticVersion`.

```suggestion
          artifactVersions.get().filterArtifactsByMajorVersion(artifactVersion.get().semanticVersion());
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -101,6 +107,23 @@ public boolean containsAnyOf(ArtifactVersions versions) {
     return false;
   }
 
+  /**
+   * Checks if the collection contains all the other versions.
+   *
+   * @param versions The other versions.
+   * @return True if all the other versions is present in the collection,
+   *          false otherwise.
```
artem-smotrakov, line 115:
```suggestion
   *         false otherwise.
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +150,36 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+   *
+   * @param version The artifact version.
+   * @return Filtered {@link ArtifactVersions}.
+   */
+  public ArtifactVersions filterArtifactsByMajorVersion(String version) {
+    Optional<SemanticVersion> currentVersion = SemanticVersion.parse(version);
+    if (!currentVersion.isPresent()) {
+      return ArtifactVersions.ofNothing();
+    }
+
```
artem-smotrakov, line 164:
```suggestion
  public ArtifactVersions filterArtifactsByMajorVersion(SemanticVersion version) {
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
```diff
@@ -127,6 +150,36 @@ public boolean containsAnyOf(ArtifactVersions versions) {
...
+    int major = currentVersion.get().getMajor();
+    ComparableVersion currentMajorVersion =
+        new ComparableVersion(String.format(MAJOR_VERSION_TEMPLATE, major));
+    ComparableVersion nextMajorVersion =
+        new ComparableVersion(String.format(MAJOR_VERSION_TEMPLATE, major + 1));
+
+    // If a non-semantic version exists, do not consider it for score calculation.
```
artem-smotrakov, line 171: This class doesn't know how artifact versions are used

```suggestion
```

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -52,8 +57,8 @@ public void testCalculateVersionStatistics() {
         new ArtifactVersion("1.0.0", now.minusDays(250))));
 
     VersionStats stats = ArtifactReleaseHistoryScore.calculateVersionStats(versionInfos);
-    Assert.assertEquals(173.33333333333334, stats.averageDaysBetweenReleases, DELTA);
-    Assert.assertEquals(0.6666666666666666, stats.releaseCycleTrend, DELTA);
+    assertEquals(173.33333333333334, stats.averageDaysBetweenReleases, DELTA);
```
artem-smotrakov, line 60: `DELTA` is only 0.2.
```suggestion
    assertEquals(173.33, stats.averageDaysBetweenReleases, DELTA);
```

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +244,141 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    Value<ArtifactVersions> versions = RELEASED_ARTIFACT_VERSIONS.value(artifactVersions);
+    Value<ArtifactVersion> version = ARTIFACT_VERSION.value(version120);
+    ScoreValue value = score.calculate(versions, version);
+    assertEquals(8.8, value.get(), DELTA);
+
+    final Optional<Value<?>> versionsValue =
+        value.usedValues().stream().filter(v -> v instanceof ArtifactVersionsValue).findFirst();
+    assertTrue(versionsValue.isPresent());
+    assertTrue(((ArtifactVersions) versionsValue.get().get()).containsAllOf(artifactVersions));
```
artem-smotrakov, line 278: Did you mean this?

```suggestion
    assertTrue(value.usedValues().contains(artifactVersions));
    assertTrue(value.usedValues().contains(version));
```

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +244,141 @@ public void testFourteenMonthOldSingleVersion() {
...
+    assertEquals(6.1, value.get(), DELTA);
+
+    final Optional<Value<?>> versionsValue =
+        value.usedValues().stream().filter(v -> v instanceof ArtifactVersionsValue).findFirst();
+    assertTrue(versionsValue.isPresent());
+    assertTrue(((ArtifactVersions) versionsValue.get().get()).containsAllOf(artifactVersions));
+  }
+
+  @Test
+  public void testVersionsSelectionIfNonSemanticVersionExists() {
```
artem-smotrakov, line 307: Isn't this test case a duplicate of `testVersionsSelectionByComparingMajor`?` 1.3.0-MIGHTY` is not a semantic version.

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +244,141 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactVersion nonSematicVersion = new ArtifactVersion("MIGHTY-1.0", now.minusDays(4 * 31));
+    ArtifactVersion version110 = new ArtifactVersion("1.1.0", now.minusDays(3 * 31));
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(2 * 31));
+
+    ArtifactVersions artifactVersions =
+        ArtifactVersions.of(version100, nonSematicVersion, version110, version200);
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    ScoreValue value = score.calculate(RELEASED_ARTIFACT_VERSIONS.value(artifactVersions),
+        ARTIFACT_VERSION.value(nonSematicVersion));
+    assertEquals(7.3, value.get(), DELTA);
```
artem-smotrakov, line 341: Could you please add a check for `2.0.0` here? The values should be same, shouldn't they?

```suggestion
    ScoreValue valueForNotSemanticVersion = score.calculate(RELEASED_ARTIFACT_VERSIONS.value(artifactVersions),
        ARTIFACT_VERSION.value(nonSematicVersion));
    assertEquals(7.3, valueForNotSemanticVersion.get(), DELTA);
    ScoreValue valueForSemanticVersion = score.calculate(RELEASED_ARTIFACT_VERSIONS.value(artifactVersions),
        ARTIFACT_VERSION.value(nonSematicVersion));
    assertEquals(7.3, valueForSemanticVersion.get(), DELTA);
```

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +244,141 @@ public void testFourteenMonthOldSingleVersion() {
...
+    LocalDateTime now = LocalDateTime.now();
+    ArtifactVersion version100 = new ArtifactVersion("1.0.0", now.minusDays(6 * 31));
+    ArtifactVersion version110 = new ArtifactVersion("1.1.0", now.minusDays(4 * 31));
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(2 * 31));
+
+    ArtifactVersions artifactVersions = ArtifactVersions.of(version100, version110, version200);
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    ScoreValue value = score.calculate(RELEASED_ARTIFACT_VERSIONS.value(artifactVersions),
+        ARTIFACT_VERSION.unknown());
+    assertEquals(5.0, value.get(), DELTA);
```
artem-smotrakov, line 378: It would be good to check if confidence is not max here. Same for other test cases that use unknown values.

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactVersionSecurityScoreTest.java
```diff
@@ -74,7 +74,7 @@ public void testCalculate() {
         PACKAGE_MANAGERS.value(PackageManagers.from(MAVEN)));
 
     ScoreValue scoreValue = score.calculate(values);
-    assertTrue(DoubleInterval.closed(7,8).contains(scoreValue.get()));
+    assertTrue(DoubleInterval.closed(6.75,7.25).contains(scoreValue.get()));
```
artem-smotrakov, line 77:
```suggestion
    assertTrue(DoubleInterval.closed(6.75, 7.25).contains(scoreValue.get()));
```

src/main/java/com/sap/oss/phosphor/fosstars/model/value/ArtifactVersions.java
sourabhsparkala: @mibo I guess we are on the same page now. Thank you for clearing this up.

mibo: @sourabhsparkala Sure and Thanks for the nice chat 😉

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
sourabhsparkala: `1.3.0-MIGHTY` is a semantic version right? I based this input from here https://github.com/SAP/fosstars-rating-core/blob/master/src/test/java/com/sap/oss/phosphor/fosstars/model/value/SemanticVersionTest.java#L27

sourabhsparkala: I agree that we need to replace `SemanticVersion`. We can do so with this with the java lib https://search.maven.org/artifact/com.vdurmont/semver4j/3.1.0/jar.

However I observed that this would completely restrict us from using versions like
- `1.2` which is a valid semantic version

And we have to be limited to versions like as given below. More information about the format [here](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions)
- `1.2.0`
- `1.2.0-SNAPSHOT` 

Then we also need to consider the possibility of using non-semantic versions. I believe we need to take this task of updating `SemanticVersion` there. I have created an issue for that #565

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +243,138 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactVersion version150 = new ArtifactVersion("1.5.0", now.minusDays(64));
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(31));
+    ArtifactVersions artifactVersions = ArtifactVersions.of(version100, version110, version120,
+        version130, version140, version150, version200);
+
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    Value<ArtifactVersions> versions = RELEASED_ARTIFACT_VERSIONS.value(artifactVersions);
+    Value<ArtifactVersion> version = ARTIFACT_VERSION.value(version120);
+    ScoreValue value = score.calculate(versions, version);
+    assertEquals(8.8, value.get(), DELTA);
```
artem-smotrakov, line 272: Could you please add `assertEquals()` for "2.0.0", "1.5.0" and "1.0.0"?

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +243,138 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactVersion version150 = new ArtifactVersion("1.5.0", now.minusDays(64));
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(31));
+    ArtifactVersions artifactVersions = ArtifactVersions.of(version100, version110, version120,
+        version130, version140, version150, version200);
+
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    Value<ArtifactVersions> versions = RELEASED_ARTIFACT_VERSIONS.value(artifactVersions);
+    Value<ArtifactVersion> version = ARTIFACT_VERSION.value(version200);
+    ScoreValue value = score.calculate(versions, version);
+    assertEquals(6.1, value.get(), DELTA);
```
artem-smotrakov, line 294: Could you please add `assertEquals()` for "1.5.0" and "1.0.0"?

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +243,138 @@ public void testFourteenMonthOldSingleVersion() {
...
+    ArtifactVersion version120 = new ArtifactVersion("1.2.0", now.minusDays(29));
+    ArtifactVersion version200 = new ArtifactVersion("2.0.0", now.minusDays(2 * 31));
+
+    ArtifactVersions artifactVersions =
+        ArtifactVersions.of(version100, nonSematicVersion, version110, version120, version200);
+    ArtifactReleaseHistoryScore score = new ArtifactReleaseHistoryScore();
+    Value<ArtifactVersions> versions = RELEASED_ARTIFACT_VERSIONS.value(artifactVersions);
+    Value<ArtifactVersion> version = ARTIFACT_VERSION.value(version110);
+    ScoreValue value = score.calculate(versions, version);
+    assertEquals(8.5, value.get(), DELTA);
```
artem-smotrakov, line 314: Could you please add `assertEquals()` for "2.0.0"?

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
```diff
@@ -40,19 +48,25 @@ public ScoreValue calculate(Value<?>... values) {
...
-    Collection<ArtifactVersion> sortedByReleaseDate =
-        ArtifactVersions.sortByReleaseDate(artifactVersions);
+    final Collection<ArtifactVersion> artifactCollection =
+        filter(artifactVersions, artifactVersion);
+
+    // If the artifact collection is filtered, then a cluster of versions have been obtained.
+    // Increase the score value.
+    if (artifactCollection.size() < artifactVersions.get().size()) {
+      scoreValue.increase(1);
+    }
```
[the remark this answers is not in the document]
sourabhsparkala: Good catch.

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
sourabhsparkala: I assumed you were talking about this `assertTrue(value.usedValues().contains(versions));`
Basically, check if the entered artifact versions are part of used values.

Now it is now clear to me. @artem-smotrakov Could you please elaborate a bit more?

artem-smotrakov: Please calculate scores for "2.0.0", "1.5.0" and "1.0.0" and check if they are expected.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
```diff
@@ -139,6 +147,30 @@ static VersionStats calculateVersionStats(Collection<VersionInfo> versionInfoCol
...
+   */
+  private static Collection<ArtifactVersion> filter(Value<ArtifactVersions> artifactVersions,
+      Value<ArtifactVersion> artifactVersion, ScoreValue scoreValue) {
+    if (!artifactVersion.isUnknown() && artifactVersion.get().hasValidSemanticVersion()) {
+      ArtifactVersions filteredArtifactVersions = artifactVersions.get()
+          .filterArtifactsByMajorVersion(artifactVersion.get().getSemanticVersion().get());
+
+      // If a cluster of versions has been obtained. Increase the score value.
+      if (filteredArtifactVersions.size() > ARTIFACT_VERSIONS_SIZE_THRESHOLD) {
+        scoreValue.increase(1);
```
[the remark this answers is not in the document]
sourabhsparkala: It has satisfied the scenarios we discussed via the test cases. Unless you had a specific test case in mind. I can try to implement that. Could you elaborate a bit more?

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
```diff
@@ -239,12 +243,154 @@ public void testFourteenMonthOldSingleVersion() {
...
+    assertEquals(8.8, value120.get(), DELTA);
+
+    ScoreValue value150 = score.calculate(versions, ARTIFACT_VERSION.value(version150));
+    assertEquals(8.8, value150.get(), DELTA);
+
+    ScoreValue value130 = score.calculate(versions, ARTIFACT_VERSION.value(version130));
+    assertEquals(8.8, value130.get(), DELTA);
+
+    ScoreValue value200 = score.calculate(versions, ARTIFACT_VERSION.value(version200));
+    assertEquals(6.1, value200.get(), DELTA);
```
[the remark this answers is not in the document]
sourabhsparkala: @artem-smotrakov Yes, it is expected.
1. If a cluster is found the rating should be specifically higher. As the score considers only the associated versions.
2. If a cluster is not found, then every version is included. This should not have a comparatively good score.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScore.java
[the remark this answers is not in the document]
artem-smotrakov: This is related to my comment for the tests https://github.com/SAP/fosstars-rating-core/pull/549#discussion_r641409349 I am not sure that the expected scores are right in the test cases. Could you please check it?

Plus, the increase happens only if the version is semantic. I don't see that semantic versions are better than any other format. Therefore, I am not sure if the increase should happen only if the version is semantic.

artem-smotrakov (approved):
I am approving this but I don't think the logic of the updated score is correct. Please see my comments.

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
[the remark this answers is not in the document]
mibo: As mentioned by Sourabh in case of a new major version release (one single version) it falls back to check all versions.
This may lead to a lower score for the latest version compared to a score to a specific major version branch.
In the case that a project release new versions fast (e.g. new version every month) also the latest (single) major version would have a good score.
In cases that e.g. 1.x branch has slow release cycle, 2.x has fast cycle (both together has an average cycle) the latest 3.x version will have an average cycle. Compared to 1.x it is better, compared to 2.x is worse. But if they continue to release for 3.x fast it will be good.

Maybe this can be improved somehow, but currently I have no idea how to do this.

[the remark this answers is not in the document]
artem-smotrakov: Thanks for considering this. I am fine with either addressing it now or later. It is up to you, I've approved the current changeset.

mibo (approved):
LGTM 👍

src/test/java/com/sap/oss/phosphor/fosstars/model/score/oss/ArtifactReleaseHistoryScoreTest.java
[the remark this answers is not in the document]
sourabhsparkala: #569

PR #555, opened 2021-05-20
HasSecurityPolicyTest.noTestPolicy() should not require access to the Internet

### LINKED ISSUE #554: HasSecurityPolicyTest.noTestPolicy() should not require access to the Internet
If no Internet connection available, the test case fails.

artem-smotrakov:
Fixes #554


sourabhsparkala (approved):
Looks good 👍

PR #560, opened 2021-05-25
OSS RoP: Add repository exceptions for license check

SebastianWolf-SAP:
Two repositories don't use standard licenses at the moment, won't change and are accepted by us. Therefore, they can be added as exceptions.

### DIFF
--- a/LicenseInfo.config.yml
+++ b/LicenseInfo.config.yml
@@ -5,3 +5,6 @@ allowedLicenses:
   - MIT
   - EPL-2.0
 disallowedLicensePatterns: "API"
+repositoryExceptions:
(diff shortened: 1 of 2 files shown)


artem-smotrakov (approved):

artem-smotrakov:
@SebastianWolf-SAP Looks good to me. Once it is merged, you can run this workflow to re-build the report ("Run workflow" button)

https://github.com/SAP/fosstars-rating-core/actions/workflows/fosstars-rules-of-play.yml

SebastianWolf-SAP:
Thx for the review. Yes, as soon as it's merged, I'll re-run the workflow.

PR #561, opened 2021-05-25
Added parameters for OSS RoP workflow

artem-smotrakov:
Docs: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch


sourabhsparkala (approved):
👍

SebastianWolf-SAP (approved):
LGTM

PR #562, opened 2021-05-26
Explanations for all values

### LINKED ISSUE #539: Values should contain explanations
Currently, a data provider creates a feature value that is then used in rating calculation. However, data providers doesn't provide any explanation how they created specific feature values. It would be good if data providers stores explanations for values. For example, the interface `Value` can have `explanations()` method that would return a text with comments from the data provider that created the value. The explanation may be then used in reports to make them more transparent.

artem-smotrakov:
This pull request addresses #539

- Added methods to read and write explanations for all values.
- Updated `LicenseInfo` data provider to explain the values.
- Updated the formatter for the OSS RoP rating to print out explanations if available.
- Updated tests.

I created #563 and #564 to make the rest of data providers explain the value they create.

Closes #539


SebastianWolf-SAP (approved):
I have to say that I don't know enough about the overall FOSSTARS architecture to assess the impact of the change (but getting there). But from what I can see here, it's really well designed and implemented and works well. Thanks for adjusting the LicenseInfo data provider already. I see that I need to adopt a little more of your coding style. seems that my Java skills are a little bit rusty... ;)

artem-smotrakov:
Thanks for the feedback @SebastianWolf-SAP !

PR #594, opened 2021-06-17
Run CLI tests in a GitHub workflow

artem-smotrakov:
Main updates:

- Added more tests for the CLI
- Added a workflow that runs tests for the CLI
- Fixed a couple of issues

Closes #589 
Closes #602

PR #592, opened 2021-06-16
Better command line tool

artem-smotrakov:
Here is a list of main updates:

- Spring the `Application` class to several handlers that run calculation of specific ratings.
- Updated the CLI to support calculating security ratings for artifacts (Maven and NPM).

Closes #585
Closes #588

PR #595, opened 2021-06-17
OSS Rules of Play gets option to create issues for findings

### LINKED ISSUE #528: OSS RoP action should create issues for violated rules
https://github.com/SAP/fosstars-rating-core/tree/master/.github/actions/oss-rules-of-play-action

This local action generates a RoP report for specified projects. It would be nice if it could open issues on GitHub for violated rules.

Things to do:
- [ ] Implement an option in the action that turns on creating issues for violated rules.
- [ ] The action should be configurable in a workflow.
- [ ] The subject of a created issue should contain the rule identifier (see in the report).
- [ ] The action should not open duplicate issues (rule identifiers may be used to find duplicates).

SebastianWolf-SAP:
This PR fixes #528. If the OSS RoP ruleset is active, the command-line option ``--create-issues`` will create issues in the given GitHub repository in case there are violations against the ruleset.
It references the rule ID in the title and the body. Moreover, a configurable URL (``documentationUrl`` in ``OssRulesOfPlayRatingMarkdownFormatter.config.yml``) is also posted along with the other information in the issue body. For an example of such an issue see https://github.com/SebastianWolf-SAP/jak-test/issues/22
Of course, it checks for already existing issues with the same title before creating new ones, so duplicates are avoided.


src/main/java/com/sap/oss/phosphor/fosstars/tool/format/OssRulesOfPlayRatingMarkdownFormatter.java
```diff
@@ -377,4 +423,24 @@ public static String format(Label label) {
...
+    JsonNode config = Yaml.mapper().readTree(Files.newBufferedReader(path));
+    if (!config.has("documentationUrl")) {
+      return StringUtils.EMPTY;
+    }
+
+    if (!config.isObject()) {
+      throw new IOException("Oops! Configuration is not an object!");
+    }
+
+    return config.get("documentationUrl").asText();
```
artem-smotrakov, line 444: It would be better to close the reader. I'll fix it later.

```suggestion
    try (BufferedReader reader = Files.newBufferedReader(path)) {
      JsonNode config = Yaml.mapper().readTree(reader);
      if (!config.has("documentationUrl")) {
        return StringUtils.EMPTY;
      }

      if (!config.isObject()) {
        throw new IOException("Oops! Configuration is not an object!");
      }

      return config.get("documentationUrl").asText();
    }
```

src/main/java/com/sap/oss/phosphor/fosstars/tool/format/OssRulesOfPlayRatingMarkdownFormatter.java
```diff
@@ -168,6 +194,25 @@ private String warningsFrom(ScoreValue scoreValue) {
...
+  
+  @Override
+  protected String formatRule(Value<?> value) {
+    return String.format("1.  **%s** %s **%s**",
+        identifierOf(value.feature()),
+        nameOf(value.feature()),
+        printValueAnswer(value));
+  }
+
+  private String printValueAnswer(Value<?> value) {
```
artem-smotrakov, line 206: Looks like this duplicates `formatRule()` a bit. We can try to get rid of dups later.

artem-smotrakov (approved):
I have a few minor comments that I'll address in #592 -- that is a big refactoring, I'll adapt your changeset, so don't be surprised if you don't find your changes in `Application` :) I am also planning to add more tests for the CLI in #594. I am not sure I can cover the new option there but at least the new tests can help to make sure that other options still work fine. I'm merging this pull request. Great job @SebastianWolf-SAP !

PR #606, opened 2021-06-23
Better OSS RoP action

### LINKED ISSUE #604: GitHub action for OSS RoP unnecessarily updates .gitignore
https://github.com/SAP/fosstars-rating-core/blob/80d7936ab38a572056e463dbdf6e2dc6f00c69f3/.github/actions/oss-rules-of-play-action/entrypoint.sh#L59

The GitHub action for OSS RoP updates `.gitignore` file when it runs. As a result, the file unnecessarily grows.

### LINKED ISSUE #605: Check if a token is available in the GitHub action for OSS RoP
https://github.com/SAP/fosstars-rating-core/blob/80d7936ab38a572056e463dbdf6e2dc6f00c69f3/.github/actions/oss-rules-of-play-action/entrypoint.sh#L49

The action expects a token but there is no check for that.

artem-smotrakov:
Closes #604
Closes #605

### DIFF
--- a/.github/actions/oss-rules-of-play-action/entrypoint.sh
+++ b/.github/actions/oss-rules-of-play-action/entrypoint.sh
@@ -62,9 +62,11 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+# Cleanup
+rm -rf fosstars-rating-core > /dev/null 2>&1
+rm -rf .fosstars > /dev/null 2>&1
+
 # Commit the report
-echo fosstars-rating-core >> .gitignore
-echo .fosstars >> .gitignore
 git add --all
 git config --global user.name "Fosstars"
 git config --global user.email "fosstars@users.noreply.github.com"
--- a/.github/actions/oss-rules-of-play-action/entrypoint.sh
+++ b/.github/actions/oss-rules-of-play-action/entrypoint.sh
@@ -15,6 +15,13 @@ if [ "$FOSSTARS_VERSION" = "" ]; then
     exit 1
 fi
 
+if [ "$TOKEN" = "" ]; then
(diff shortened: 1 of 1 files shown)


.github/actions/oss-rules-of-play-action/entrypoint.sh
```diff
@@ -15,6 +15,13 @@ if [ "$FOSSTARS_VERSION" = "" ]; then
     exit 1
 fi
 
+if [ "$TOKEN" = "" ]; then
+    echo "Achtung! No token provided or it is empty!"
+    TOKEN_OPTION=""
```
sourabhsparkala, line 20: Do we exit if there is no token?

sourabhsparkala (approved):
A minor comment

.github/actions/oss-rules-of-play-action/entrypoint.sh
artem-smotrakov: No, the action will continue to run but without token.

.github/actions/oss-rules-of-play-action/entrypoint.sh
```diff
@@ -55,9 +62,11 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+# Cleanup
+rm -rf fosstars-rating-core > /dev/null 2>&1
+rm -rf .fosstars > /dev/null 2>&1
+
```
sourabhsparkala, line 68: Could this be affected by some folder permissions?

artem-smotrakov: Those directories are created by the action (by same process/user that tries to delete them), so there should not be issues with permissions.

PR #620, opened 2021-07-01
Better explanations for RoP

### LINKED ISSUE #563: Data providers should explain values for the OSS RoP rating
#539 introduced explanations for all values. The data providers, which gather data for the OSS RoP rating, should explain the values they create.

### LINKED ISSUE #619: Better advice selection in OssRulesOfPlayAdvisor
https://github.com/SAP/fosstars-rating-core/blob/18697b2ee67ca6d782afca13da7e6ba9530e0fad/src/main/java/com/sap/oss/phosphor/fosstars/advice/oss/OssRulesOfPlayAdvisor.java#L84

`OssRulesOfPlayAdvisor` uses `AbstractOssAdvisor.knownFalseValue()` predicate to select advice for values. That is wrong because some values are expected to be false. As a result, advice may be displayed for a passed rule.

### LINKED ISSUE #622: ReadmeInfo should load default config
The data provider `ReadmeInfo` does not look for a default config.

artem-smotrakov:

- Updated the data providers to Data add explanations for values. This closes #563
- Introduced a draft of builders for building Markdown reports. I noticed that the number of nested if/else/switch/for/etc blocks is growing in the formatters. The builders won't probably significantly reduce the total number of the block, but there should be less nested blocks. Furthermore, it is going to be easier to add unit tests. Currently, the formatters are barely covered by unit tests. This work is going to be continued in a follow-up #621 . The builders will probably move to separate package/classes, and if so, I'll add javadoc for them.
- Updated `OssRulesOfPlayRatingMarkdownFormatter` to print both explanations and advice in the same section.
- Fixed #619
- Fixed #622

[Example of new report](https://github.com/artem-smotrakov/fosstars-rating-core/blob/fosstars-oss-rules-of-play-test-report/artem-smotrakov/test-rop-one.md).


sourabhsparkala (approved):
Looks good to me 👍

PR #623, opened 2021-07-05
Update LicenseInfo.config.yml

ajinkyapatil8190:
Repositories with exceptional licenses, need to be entered to the exception list for the Repository Linter and FOSSTARS:
https://github.com/SAP/abnf-test-tool
https://github.com/SAP/software-documentation-data-set-for-machine-translation


artem-smotrakov (commented):
Looks fine to me. @SebastianWolf-SAP Could you please also have a look at it?

SebastianWolf-SAP (approved):
Back from vacation... LGTM as well - as agreed these exceptions need to be added.

PR #625, opened 2021-07-06
Updated weights and simplify test vectors

### LINKED ISSUE #616: Revisit score weights in the OSS security score
It's been a while since the weights were configured in the OSS security score. Let's review them and see if it makes sense to update them.

### LINKED ISSUE #624: Simplify test vectors for VulnerabilityDiscoveryAndSecurityTestingScore
Test vectors for `VulnerabilityDiscoveryAndSecurityTestingScore` contains full score values that contain a lot of unnecessary details.

artem-smotrakov:
Here are main updates:

- Store scores in YAML. It allows adding comments for weights.
- Updated weights for the OSS security score.
- Simplified test vectors for `VulnerabilityDiscoveryAndSecurityTestingScore`.

Closes #616
Closes #624


src/main/java/com/sap/oss/phosphor/fosstars/model/qa/ScoreTestVector.java
```diff
@@ -69,7 +72,10 @@ public ScoreTestVector(
 
     Set<Value<?>> result = new HashSet<>();
     for (Map.Entry<Class<? extends Score>, Double> entry : values.entrySet()) {
-      Score subScore = subScore(score, entry.getKey());
+      String targetScoreClassName = entry.getKey().getCanonicalName();
+      Score subScore = subScoreIn(score, targetScoreClassName)
+          .orElseThrow(() -> new IllegalArgumentException(
+              format("Could not fine sub-score %s!", targetScoreClassName)));
```
sourabhsparkala, line 78: typo

src/main/resources/com/sap/oss/phosphor/fosstars/model/score/oss/OssSecurityScoreWeights.yml
```diff
@@ -0,0 +1,26 @@
...
+    value: 0.9
+  com.sap.oss.phosphor.fosstars.model.score.oss.UnpatchedVulnerabilitiesScore:
+    type: ImmutableWeight
+    value: 0.5 # unfortunately, we don't have a reliable source of info about unpatched vulnerabilities
+  com.sap.oss.phosphor.fosstars.model.score.oss.ProjectPopularityScore:
+    type: ImmutableWeight
+    value: 0.5
+  com.sap.oss.phosphor.fosstars.model.score.oss.ProjectActivityScore:
+    type: ImmutableWeight
+    value: 0.5
```
sourabhsparkala, line 14: I think these scores can have better weights
- The reason being, the data which can be actually crawled are used to calculate these scores. Yes? 
- Shouldn't the weight justify depending on its availability?

sourabhsparkala (approved):
Looks good. 
The weights assigned. Do we need some sort of justification with a doc? as to why these weights were assigned to the scores?

src/main/resources/com/sap/oss/phosphor/fosstars/model/score/oss/OssSecurityScoreWeights.yml
artem-smotrakov: I am not sure that I understand these two points. How exactly you would you update the weights?

artem-smotrakov:
> The weights assigned. Do we need some sort of justification with a doc? as to why these weights were assigned to the scores?

Currently, we don't explain the weights in the docs. If you think we need it, please open an issue.

sourabhsparkala (approved):
👍

src/main/resources/com/sap/oss/phosphor/fosstars/model/score/oss/OssSecurityScoreWeights.yml
sourabhsparkala: Well, my understanding is this:
- If there are few Data Providers which we are almost absolutely sure that we can get some data. May the scores generated using that data should have a higher weight.
- For example. we know we can get no of contributors of a github project, but LGTM, CodeQL related data providers may have no data.

artem-smotrakov: I'd use such knowledge to reduce weight if we know that data is not very reliable or complete. But possible upper bounds should not be derived from that. If we tune the security score, I believe we should think first what kind of scores tell about project's security the most. For example, if we have more reliable data about project activity, its weight should not be higher than the weight for security testing.

sourabhsparkala (approved):

PR #629, opened 2021-07-13
Better Markdown builders

### LINKED ISSUE #621: Better Markdown builders
#620 introduced a draft of builders for Markdown. Here are possible further improvements:

- [x] Consider moving the builders to separate package/classes.
- [x] Add more elements and builders.
- [x] Add more tests.
- [x] Add javadoc.
- [x] Update other Markdown formatters to use the builders.

artem-smotrakov:
Updated the existing Markdown formatters to use Markdown builders.

Closes #621


src/main/java/com/sap/oss/phosphor/fosstars/tool/format/AbstractMarkdownFormatter.java
```diff
@@ -350,14 +332,15 @@ String detailsOf(ScoreValue scoreValue) {
...
           subScoreValues.size() == 1 ? StringUtils.EMPTY : "s"));
-      sb.append(highLevelDescriptionOf(scoreValue));
-      sb.append("\n");
+
+      elements.add(highLevelDescriptionOf(scoreValue));
     }
 
     if (!featureValues.isEmpty()) {
-      sb.append(format("This sub-score is based on %d feature%s:%n%n",
+      elements.add(Markdown.string("This sub-score is based on %d feature%s:%n%n",
```
sourabhsparkala, line 343: Make these `Markdown.string` as static import?

sourabhsparkala (approved):
:+1

One minor comment.

src/main/java/com/sap/oss/phosphor/fosstars/tool/format/AbstractMarkdownFormatter.java
artem-smotrakov: I thought about it but `Markdown.string` sounds better to me. Let's keep it as-is. If we need to make the lines shorter in the future, we can add a static import.

artem-smotrakov: I thought about it but `Markdown.string` sounds better to me. Let's keep it as-is. If we need to make the lines shorter in the future, we can add a static import.

artem-smotrakov:
Thanks for the review @sourabhsparkala I'll merge it after you release `1.4.0`.

PR #640, opened 2021-07-19
RoP: Follow symbolic links for REUSE licenses

### LINKED ISSUE #639: RoP: REUSE check ignores symlink entries
Though the repository is compliant and references licenses in the expected format, the reuse data provider does not interpret the content correctly if symlinks are used, e.g. https://github.com/SAP/fosstars-oss-rules-of-play-report/blob/main/SAP/go-ase.md

SebastianWolf-SAP:
This one fixes #639 (without explicit test, but that might be really hard to achieve). However, you probably had your reasons why you explicitly disabled following symlinks in the original commit. So in case you have certain reasons in mind to keep it this way, I'm open to a chat about it. :)


artem-smotrakov (approved):
I don't remember the reason why symlinks should be skipped. Let's give it a try.

PR #641, opened 2021-07-20
Added hashCode() for a few base value classes

### LINKED ISSUE #627: LGTM: Inconsistent equals and hashCode
LGTM reports a few alerts about `equals()` and `hashCode()`

https://lgtm.com/projects/g/SAP/fosstars-rating-core/?mode=tree&ruleFocus=6770060

artem-smotrakov:
In fact, I don't think that the LGTM findings are valid. In this case, it is okay not to implement `hashCode()`, it is implemented in `AbstractValue`. Anyway, I added `hashCode()` methods to make LGTM happy. I also made several methods final.

Fixes #627


src/main/java/com/sap/oss/phosphor/fosstars/model/value/AbstractKnownValue.java
```diff
@@ -62,4 +62,9 @@ public boolean equals(Object o) {
     }
     return super.equals(o);
   }
+
+  @Override
+  public int hashCode() {
+    return 17 * super.hashCode();
```
sourabhsparkala, line 68: Any specific reason for 17 here?

sourabhsparkala (approved):
👍

src/main/java/com/sap/oss/phosphor/fosstars/model/value/AbstractKnownValue.java
artem-smotrakov: No.

PR #651, opened 2021-07-23
Use number of dependents in the project popularity score

### LINKED ISSUE #630: Project popularity score: 3.38 for jackson-databind
Though Jackson-databind is quite popular and widely used, we get the the Project popularity score as 3.38 for the same with the below details:

```
Project popularity
Score: 3.38, confidence is 10.0 (max), weight is 0.5 (medium)

The score is based on number of stars and watchers. Here is how a number of stars contributes to the score: 0 -> 0.00 (min), 2500 -> 2.50, 5000 -> 5.00, 10000 -> 10.00 (max) Here is how a number of watchers contributes to the score: 0 -> 0.00 (min), 450 -> 1.50, 750 -> 2.50, 3000 -> 10.00 (max)

This sub-score is based on 2 features:

Number of stars for a GitHub repository: 2858
Number of watchers for a GitHub repository: 158
```

I would expect some 8 to 9 as the Project popularity score for it as its widely used and known in Open source community.

artem-smotrakov:
Here are main updates:

- Added `NUMBER_OF_DEPENDENT_PROJECTS_ON_GITHUB` feature.
- Added `NumberOfDependentProjectOnGitHub` provider.
- Updated the popularity score to use the new feature.

Closes #630

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ProjectPopularityScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ProjectPopularityScore.java
@@ -1,24 +1,21 @@
 package com.sap.oss.phosphor.fosstars.model.score.oss;
 
+import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.NUMBER_OF_DEPENDENT_PROJECTS_ON_GITHUB;
 import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.NUMBER_OF_GITHUB_STARS;
 import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.NUMBER_OF_WATCHERS_ON_GITHUB;
 import static com.sap.oss.phosphor.fosstars.model.other.Utils.findValue;
 
 import com.sap.oss.phosphor.fosstars.model.Value;
-import com.sap.oss.phosphor.fosstars.model.math.MathHelper;
 import com.sap.oss.phosphor.fosstars.model.score.FeatureBasedScore;
 import com.sap.oss.phosphor.fosstars.model.value.ScoreValue;
-import java.util.function.Function;
 
 /**
  * <p>The project popularity score is currently based on two features.</p>
  * <ul>
  *   <li>Number of stars on GitHub</li>
  *   <li>Number of watchers on GitHub</li>
+ *   <li>Number of projects that use the project</li>
  * </ul>
- * <p>First, it calculates a stars sub-score. Next, it calculates a watchers sub-score.
- * It uses linear functions to transform the numbers to sub-scores.
- * Then, the sub-scores are added to each other.</p>
  */
 public class ProjectPopularityScore extends FeatureBasedScore {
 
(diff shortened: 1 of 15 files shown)


src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb
```diff
@@ -4390,7 +4390,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.1"
+   "version": "3.6.9"
```
sourabhsparkala, line 4393: Was there any issue in python version 3.7.1?

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ProjectPopularityScore.java
```diff
@@ -27,114 +24,67 @@
...
-      throw new IllegalArgumentException("Number of watchers is negative! How can it be possible?");
+    int n = value.get();
+    if (n < 0) {
+      throw new IllegalArgumentException("Oops! Expected a non-negative value!");
     }
 
-    if (m < BEST_WATCHERS_AMOUNT) {
-      return WATCHERS_SCORE_FACTOR * m;
+    if (n < threshold) {
+      return MAX / threshold * n;
```
sourabhsparkala, line 87: If n = 0? Could this lead to some `ArithmeticException`?

sourabhsparkala: It should not ideally as you are multiplying n here instead of `n/threshold`. I think it should be fine.

sourabhsparkala (dismissed):
Minor comments.

However if there are any UI changes from Github. Could this data provider fail?

src/main/jupyter/oss/security/SecurityRatingAnalysis.ipynb
artem-smotrakov: I didn't notice any issue. Actually I didn't update this field. Looks like Jupyter updated it. To be honest, I am really familiar with the structure of `ipynb` files. Do you think this update can cause some problems?

artem-smotrakov: I'll just revert this update. I think I updated the notebook in #652 and committed this update here by mistake.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ProjectPopularityScore.java
artem-smotrakov: > If n = 0? Could this lead to some ArithmeticException?

If n = 0, then `MAX / threshold * n` is 0. I can update it to `MAX * (n / threshold)` if it makes it clearer. The result will be the same.

artem-smotrakov:
> Minor comments.

I've addressed them. Thanks!

> However if there are any UI changes from Github. Could this data provider fail?

Yes, the provider depends on the UI. It is not good for sure. Unfortunately, I couldn't get the number of dependent projects neither with the classic GitHub API nor GraphQL API. If they add API that allows fetching this data, then we'll need to update the provider to use this API. Meanwhile, let's keep an eye on amount of unknown values for the new feature. I'll add the feature to the notebook.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/ProjectPopularityScore.java
sourabhsparkala: Apologies, my mistake. I believe what you did early was correct. `MAX/threshold * n`. This way it will avoid this `ArithmaticException`
https://www.baeldung.com/java-division-by-zero#integers

Because before this `if` check, you check if `n < 0`

sourabhsparkala: @artem-smotrakov do you agree?

artem-smotrakov: The previous expression `MAX / threshold * n` and the current one `MAX * n / threshold` do the same and don't throw exceptions because `threshold` can't be currently 0. Neither of these two expressions divide by `n`.

sourabhsparkala (approved):
ok

PR #654, opened 2021-07-27
Don't load data from NVD for RoP

### LINKED ISSUE #653: Don't load data from NVD for RoP
When the CLI is run with a config for RoP, it still loads NVD although RoP doesn't need it. The data require quite a lot of memory though.

artem-smotrakov:
Fixes #653


SebastianWolf-SAP (approved):
Makes perfect sense, let's see if it works better with that...

PR #659, opened 2021-07-30
Use ROP_TOKEN in the fosstars-rop-report.yml workflow

artem-smotrakov:
Although the Fosstars action successfully finished and created a RoP report, it failed to run several check, see exceptions in the logs:

https://github.com/SAP/fosstars-rating-core/runs/3197786874?check_suite_focus=true

It happened because `GITHUB_TOKEN` doesn't have enough permissions to read into about teams, etc.

I updated the workflow to use `ROP_TOKEN` that currently contains my personal token (it expires in 30 days). Let's see whether it is enough or not.


SebastianWolf-SAP (approved):
That's really an issue - I was also thinking about this when I finished the work on integrating the new rating [in the GH Action](https://github.com/SAP/fosstars-rating-core-action)... We should [at least document that](https://github.com/SAP/fosstars-rating-core/issues/661), but this change is of course OK.

PR #662, opened 2021-08-03
Fail gracefully when issues are disabled

### LINKED ISSUE #660: Issue creation should check if issues are enabled
The new issue creation functionality of course doesn't work if issues are disabled for the respective repository. Instead of trying to create issues for findings, it should check the settings beforehand to avoid error messages in the log.

Errors:
```
[!] Oops! Could not create issues for https://github.com/org/repo
org.kohsuke.github.GHFileNotFoundException: https://api.github.com/repos/org/repo/issues {"message":"Issues are disabled for this repo","documentation_url":"https://docs.github.com/v3/issues/"}
```

SebastianWolf-SAP:
Fixes #660 

Also introduces two small unit tests to verify that the issue creation is invoked correctly.

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kohsuke.github.GHFileNotFoundException;
 import org.kohsuke.github.GHIssue;
 
 /**
@@ -86,8 +87,13 @@ public class OssRulesOfPlayGitHubIssuesReporter implements Reporter<GitHubProjec
       String issueHeader = printTitle(violation);
       List<GHIssue> existingGitHubIssues = fetcher.gitHubIssuesFor(project, issueHeader);
       if (existingGitHubIssues.isEmpty()) {
-        fetcher.createGitHubIssue(project, printTitle(violation), printBody(violation));
-        LOGGER.info("New issue: " + issueHeader);
+        try {
+          fetcher.createGitHubIssue(project, printTitle(violation), printBody(violation));
+          LOGGER.info("New issue: " + issueHeader);
+        } catch (GHFileNotFoundException issuesDisabledException) {
+          LOGGER.warn("Creating issues is disabled for project {}", project.toString());
+          break;
+        }
       } else {
         LOGGER.info("Issue already exists: " + issueHeader);
       }
--- a/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
@@ -13,7 +13,6 @@ import java.io.IOException;
 import java.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.kohsuke.github.GHFileNotFoundException;
 import org.kohsuke.github.GHIssue;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
```diff
@@ -86,8 +87,13 @@ public void createIssuesFor(GitHubProject project) throws IOException {
...
       if (existingGitHubIssues.isEmpty()) {
-        fetcher.createGitHubIssue(project, printTitle(violation), printBody(violation));
-        LOGGER.info("New issue: " + issueHeader);
+        try {
+          fetcher.createGitHubIssue(project, printTitle(violation), printBody(violation));
+          LOGGER.info("New issue: " + issueHeader);
+        } catch (GHFileNotFoundException issuesDisabledException) {
+          LOGGER.warn("Creating issues is disabled for project {}", project.toString());
+          break;
+        }
```
artem-smotrakov, line 96: I am not sure that `GHFileNotFoundException` means only that issues are disabled.

To check whether issues are enabled or not, you can use `hasIssues()` method

https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHRepository.html#hasIssues()
https://docs.github.com/en/rest/reference/repos#create-an-organization-repository--parameters

Then, you don't need to break the loop if an exception occurred due to some other problems.

```suggestion
        } catch (GHFileNotFoundException e) {
          LOGGER.warn("Could not create an issue for project {}", project.toString(), e);
        }
```

src/test/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporterTest.java
```diff
@@ -47,4 +63,59 @@ public void testPrintTitleAndBody() throws IOException {
...
+          new OssRulesOfPlayRatingMarkdownFormatter(new OssRulesOfPlayAdvisor()));
+
+      when(ghDataFetcher.gitHubIssuesFor(any(), any())).thenReturn(Lists.newArrayList());
+      when(ghDataFetcher.createGitHubIssue(any(), any(), any())).thenThrow(
+          GHFileNotFoundException.class);
+
+      reporter.createIssuesFor(project);
+      verify(ghDataFetcher, times(1)).createGitHubIssue(any(), any(), any());
+    } finally {
+      FileUtils.forceDeleteOnExit(CONFIG_PATH.toFile());
```
artem-smotrakov, line 90: The test case doesn't seem to create this file.

SebastianWolf-SAP: Indeed - missed to delete it after copy/pasting, will be corrected, thanks!

src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
SebastianWolf-SAP: Of course, I'm not 100% sure about the mapping of the response codes with the exceptions as I haven't digged deeper into the code. However, according to the response codes I'd say that this exception might only occur in case of a 404 or 410 (the latter [occurs if creating issues is disabled](https://docs.github.com/en/rest/reference/issues#create-an-issue) so I was confident enough to use it here.

Moreover, due to the rate limitation, I tried to avoid every GH API possible - that's why I used the try/catch approach instead of the `hasIssues()` method. But probably getting the repository information doesn't count against the issue creation numbers and is probably cached anyway, so the effect is minimal. Therefore, I don't object changing it accordingly. Expect a new commit shortly. :)

artem-smotrakov (approved):
I thought you're going to keep the try-catch block and add a `hasIssues()` check. The try-catch block could actually help when an error occurs due to some other temporarily issue. But I'll leave it up to you. Please feed free to merge it as-is, or add the try-catch block back.

SebastianWolf-SAP:
Well, if there is really another, probably temporary issue, I'd rather keep it as is as the overall process continues nicely even if an IOException (of whatever source) occurs. So, I'd merge it as soon as the automated checks have passed.

artem-smotrakov:
> Well, if there is really another, probably temporary issue, I'd rather keep it as is as the overall process continues nicely even if an IOException (of whatever source) occurs.

I mean if there are 10 issues to be created for one project, and when an exception occurs while creating the first issue, the rest 9 won't be created (well, attempted to be created) without that try-catch block. It doesn't looks like a big issue though -- we can rust rerun the job.

> So, I'd merge it as soon as the automated checks have passed.

No problem.

PR #668, opened 2021-08-03
Added --pom option for CLI

### LINKED ISSUE #345: Add --pom parameter to the command line tool
It would be good if the command line tool had `--pom` options that accepts a path to a `pom.xml` file, searches for dependencies, and calculate security ratings for them. This may be helpful in implementing a Maven plugin, see #344

artem-smotrakov:
Closes #345


sourabhsparkala (approved):
👍

PR #673, opened 2021-08-09
OSS security scoring function should return an unknown value for all …

### LINKED ISSUE #672: OSS security scoring function should return an unknown value for all unknown features
Currently, `OssSecurityScore` returns a known value if all features are unknown. All underlying scoring functions needs to be checked that they return an unknown value if all features are unknown.

artem-smotrakov:
Updates:

- Updated multiple scoring functions to return an unknown value if all features are unknown.

Fixes #672


sourabhsparkala (approved):
👍

PR #676, opened 2021-08-13
GitHub issue creation: Use identifer only for existence check

SebastianWolf-SAP:
Our regular RoP check runs lead to duplicate issues (see e.g. https://github.com/SAP/styleguides/issues/226 and https://github.com/SAP/styleguides/issues/207) because we don't seem to check the issue identifier only, but use the complete title. As the other issues might have the same ID, but different title texts, the detection fails. This PR fixes that.

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
@@ -91,7 +91,7 @@ public class OssRulesOfPlayGitHubIssuesReporter implements Reporter<GitHubProjec
       String issueHeader = printTitle(violation);
       List<GHIssue> existingGitHubIssues = fetcher.gitHubIssuesFor(project, issueHeader.substring(issueHeader.indexOf("["), issueHeader.indexOf("]") + 1));
       if (existingGitHubIssues.isEmpty()) {
-        fetcher.createGitHubIssue(project, printTitle(violation), printBody(violation));
+        fetcher.createGitHubIssue(project, issueHeader, printBody(violation));
         LOGGER.info("New issue: " + issueHeader);
       } else {
         LOGGER.info("Issue already exists: " + issueHeader);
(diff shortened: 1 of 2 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/tool/report/OssRulesOfPlayGitHubIssuesReporter.java
```diff
@@ -89,9 +89,10 @@ public void createIssuesFor(GitHubProject project) throws IOException {
 
     for (Value<Boolean> violation : violations) {
       String issueHeader = printTitle(violation);
-      List<GHIssue> existingGitHubIssues = fetcher.gitHubIssuesFor(project, issueHeader);
+      List<GHIssue> existingGitHubIssues = fetcher.gitHubIssuesFor(project,
+          issueHeader.substring(issueHeader.indexOf("["), issueHeader.indexOf("]") + 1));
```
artem-smotrakov, line 93:
```suggestion
      String withIssueTitle = formatter.identifierOf(violation.feature()).orElse(issueHeader);
      List<GHIssue> existingGitHubIssues = fetcher.gitHubIssuesFor(project, withIssueTitle);
```

artem-smotrakov (approved):
Looks good to me. Looks like the current version can throw a `IndexOutOfBoundsException` if no config with rule IDs provided. To avoid that, you can also get a rule ID  from the formatter.

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#substring(int,%20int)

But I am okay to merge the current version -- please feel free to do that.

SebastianWolf-SAP:
> Looks good to me. Looks like the current version can throw a `IndexOutOfBoundsException` if no config with rule IDs provided. To avoid that, you can also get a rule ID from the formatter.
> 
> https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#substring(int,%20int)
> 
> But I am okay to merge the current version -- please feel free to do that.

Well, the correction looks better. :) If the build runs through, I'll merge it. Thanks!

artem-smotrakov (approved):

PR #680, opened 2021-08-16
RoP documentation: Add explanation about licenses

SebastianWolf-SAP:
The duplication of license files in the root folder and the LICENSES folder often leads to confusion. The rationale is probably not sufficiently described in the documentation yet. This PR adds the necessary background information and also includes a workaround to about duplicate physical files.


artem-smotrakov (approved):
Looks good to me.

PR #682, opened 2021-08-18
RoP: Also search for lowercase READMEs

### LINKED ISSUE #681: RoP: Readme existence check ignores lower-case files
The `ReadmeInfo` data provider currently seems to ignore lowercase readme files though they are detected by GitHub properly, see e.g. https://github.com/SAP/apibusinesshub-integration-recipes and the corresponding issue https://github.com/SAP/apibusinesshub-integration-recipes/issues/43

SebastianWolf-SAP:
Fixes #681. 

However, still doesn't capture cases like 'ReadMe.md', but eventually we should migrate to [GitHub's Readme API](https://docs.github.com/en/rest/reference/repos#get-a-repository-readme) anyway. Let's see how it goes... ;)

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/data/github/ReadmeInfo.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/data/github/ReadmeInfo.java
@@ -38,7 +38,8 @@ public class ReadmeInfo extends GitHubCachingDataProvider {
    * A list of known README file names.
    */
   private static final List<String> KNOWN_README_FILES
-      = Arrays.asList("README", "README.txt", "README.md", "README.rst");
+      = Arrays.asList("README", "README.txt", "README.md", "README.rst",
+          "readme", "readme.txt", "readme.md", "readme.rst");
 
   /**
    * A list of patterns that describe required content in README.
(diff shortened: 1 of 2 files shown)


artem-smotrakov (approved):
Looks good to me. 

Since README is supposed to be in the root,`readmeIn()` could probably look for all files in the root and compare the file names with `KNOWN_README_FILES` using lowercase. That would probably fix this:

> However, still doesn't capture cases like 'ReadMe.md', but eventually we should migrate to GitHub's Readme API anyway. Let's see how it goes... ;)

Using the GitHub API would probably fix it as well. Yeah, let's see how it goes.

SebastianWolf-SAP:
> Looks good to me.
> 
> Since README is supposed to be in the root,`readmeIn()` could probably look for all files in the root and compare the file names with `KNOWN_README_FILES` using lowercase. That would probably fix this:
> 
> > However, still doesn't capture cases like 'ReadMe.md', but eventually we should migrate to GitHub's Readme API anyway. Let's see how it goes... ;)

Also thought about this solution, but well - I'll take care of it as soon as we have a real-life issue for that.

> Using the GitHub API would probably fix it as well. Yeah, let's see how it goes.

Precisely. Thanks for the review, I'll merge it then...

PR #762, opened 2021-12-23
Implement experimental Bandit scan check

### LINKED ISSUE #729: PoC: Bandit in Python projects
- To setup Bandit as part of the evaluation step, we need to setup a pre-commit configuration.
- As stated here https://github.com/PyCQA/bandit#version-control-integration, we can parse through the config.yml and check if Bandit is part of the project setup.
- Existence of `.bandit` file https://github.com/PyCQA/bandit#per-project-command-line-args
- Existence of a badge https://github.com/PyCQA/bandit#show-your-style
- A github action is present for bandit https://github.com/marketplace/actions/bandit-check

DOD:
- The above things are done.

sourabhsparkala:
Fixes #729


SebastianWolf-SAP (commented):
No detailed knowledge about Bandit and also not that much time to dig deeper here, so unfortunately I can't give an explicit approval. 
However, code looks very good, I agree with the weighting of the Bandit score, also the test seem to cover the most relevant aspects. As the checks ran through properly, it should be OK...

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * The data provider gathers info about how a project uses Bandit for static analysis. In
+ * particular, it tries to fill out the following features:
+ * <ul>
+ *   <li>{@link OssFeatures#RUNS_BANDIT_SCANS}</li>
+ *   <li>{@link OssFeatures#USES_BANDIT_SCAN_CHECKS}</li>
+ * </ul>
+ */
+public class BanditDataProvider extends GitHubCachingDataProvider {
```
artem-smotrakov, line 36: I guess this provider is based on some existing one related to CodeQL. There may be opportunities to reduce duplicate code.

src/main/java/com/sap/oss/phosphor/fosstars/model/feature/oss/OssFeatures.java
```diff
@@ -222,6 +222,22 @@ private OssFeatures() {
   public static final Feature<Boolean> USES_CODEQL_CHECKS
       = new BooleanFeature("If a project runs CodeQL checks for commits");
 
+  /**
+   * Shows if an open-source project runs Bandit scans.
+   *
+   * @see <a href="https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsrun">GitHub workflow action job config to run Bandit code scanning for a repository</a>
```
artem-smotrakov, line 228: I'd use a link to Bandit instead of GitHub Actions.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
```diff
@@ -0,0 +1,187 @@
...
+    super(fetcher);
+  }
+
+  @Override
+  public Set<Feature<?>> supportedFeatures() {
+    return setOf(RUNS_BANDIT_SCANS, USES_BANDIT_SCAN_CHECKS);
+  }
+
+  @Override
+  protected ValueSet fetchValuesFor(GitHubProject project) throws IOException {
```
artem-smotrakov, line 70: Bandit can be run without GitHub actions. This data provider may be updated to catch it.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/BanditScore.java
```diff
@@ -0,0 +1,83 @@
...
+
+    ScoreValue scoreValue = scoreValue(MIN,
+        usesBanditChecks, runsBanditScans, languages);
+
+    if (allUnknown(usesBanditChecks, runsBanditScans, languages)) {
+      return scoreValue.makeUnknown().explain(
+          "The score value is unknown because all required features are unknown.");
+    }
+
+    if (!languages.isUnknown() && !SUPPORTED_LANGUAGES.containsAnyOf(languages.get())) {
```
artem-smotrakov, line 68: If the languages feature is unknown, but the rest two features are false, then the scoring function produces 0 which may not be always correct. In this case, you may consider returning N/A but I am not sure though.

src/test/java/com/sap/oss/phosphor/fosstars/data/github/CodeqlDataProviderTest.java
```diff
@@ -97,7 +97,7 @@ private void testCodeqlRuns(String filename, InputStream content, Value<?>... ex
     CodeqlDataProvider provider = new CodeqlDataProvider(fetcher);
     ValueSet values = provider.fetchValuesFor(PROJECT);
 
-    assertEquals(2, expectedValues.length);
+    assertEquals(2, values.size());
```
artem-smotrakov, line 100: I guess, this update is not necessary.

artem-smotrakov (approved):
I've left several comments you may want to consider. Otherwise, looks fine. I can't officially approve the PR anymore because I am not in the org.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: #772 772

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: #772

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/BanditScore.java
sourabhsparkala: It could also be that 
- The languages data provider fails to collect the language information but the project still triggers the Bandit scan. 
- I guess it is up to us to decide how we want to take this up.

src/test/java/com/sap/oss/phosphor/fosstars/data/github/CodeqlDataProviderTest.java
sourabhsparkala: There is already a Test confirming that the possible features that will be available here https://github.com/SAP/fosstars-rating-core/pull/762/files/f9d5d97620fce7554c112f9a58b257816e56aa96#diff-7567ba8d1e196c5107745af1dcfa55c3f64d94c9b85201c05247500f2c7944a8L59

But you are right. As we are ourselves passing the expected values. The test is the same.

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/BanditScore.java
sourabhsparkala: I have added a check for `unknown` language. I decided that it doesn't make sense to check for a Bandit scan if there is no mention that the project uses Python scripts.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: Refactored it :)

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
```diff
@@ -0,0 +1,119 @@
...
+ *   <li>{@link OssFeatures#USES_BANDIT_SCAN_CHECKS}</li>
+ * </ul>
+ */
+public class BanditDataProvider extends AbstractStaticScanToolsDataProvider {
+
+  /**
+   * A step in a GitHub action that triggers analysis with Bandit.
+   */
+  private static final Pattern RUN_STEP_BANDIT_REGEX_PATTERN
+      = Pattern.compile("^.*bandit .*$", Pattern.DOTALL);
```
artem-smotrakov, line 37: I guess, the whitespace is here for a reason.

artem-smotrakov (approved):
I had a quick look, looks fine. I am not sure whether you really need that new interface. I think the most important thing is to detect Bandit not only in GitHub actions. I guess you're going to address it separately.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: yes, the syntax goes `bandit -r`. Just to indicate that bandit is a function call during `run:` in GitHub actions and not randomly occuring command word. Maybe I should include all the parameters possible which comes with bandit run.

However to improve this, let me work on this here #789 789

PR #773, opened 2022-01-21
Addition of the CodeQL scans to OSS Rules of Play.

ajinkyapatil8190:
This PR - 
Creates checks for code scan alerts on a repository
Recommends correct action if the code scans are not active.
Solves #775


src/test/shell/tool/github/OssRulesOfPlayRatingMarkdownFormatter.config.yml
```diff
@@ -20,3 +20,5 @@ ruleIds:
   rl-security_policy-1: If an open-source project has a security policy
   rl-code_of_conduct_file-1: If a project has a code of conduct file
   rl-code_of_conduct_file-2: If a project's code of conduct guideline has required text
+  rl-codescan_alerts-1: If a project has code scan enabled
```
sourabhsparkala, line 23: The content here should be `If a project runs CodeQL scans`

src/test/shell/tool/github/OssRulesOfPlayRatingMarkdownFormatter.config.yml
```diff
@@ -20,3 +20,5 @@ ruleIds:
...
+  rl-codescan_alerts-2: If a project has code scan on pull requests enabled
```
sourabhsparkala, line 24: Here `If a project runs CodeQL checks for commits`

ajinkyapatil8190: Changed it to "Does it run CodeQL scans?" and " Does it use CodeQL checks for pull requests? " to match the mention in other places in the code .

docs/getting_oss_rules_of_play_rating.md
```diff
@@ -23,6 +23,10 @@ The data provider `CodeOfConductGuidelineInfo` allows you to specify 1 type of r
 - The content that you expect to be part of the code of conduct guidelines.
 The option is called `requiredContentPatterns` and expects a list of one Regex. If you enter multiple entries, the code of conduct guidelines must comply with all the patterns for the rule to pass.
 
+### Code Scan Alerts
+The data provider `CodeqlDataProvider` allows you to check if the code scans are enabled for the project or not.
+It also checks if the code scans are availble on pull requests or not.
```
sourabhsparkala, line 28: We we are checking for a specific code scan `CodeQL` please be specific. A reference URL is always good. Please rephrase this text about the use case a bit better? Maybe with bullet points as there are 2 things that the CodeqlDataProvider collects.

docs/getting_oss_rules_of_play_rating.md
```diff
@@ -79,11 +83,11 @@ If everything is correctly set up, the output is going to look like the followin
 
 ```
 [+] Okay, we have a GitHub token, let's try to use it
-[+] The default value cache doesn't exist yet: .fosstars/github_project_value_cache.json
```
sourabhsparkala, line 82: It looks like you ran the build with already cached values. Could you please remove this file and the cache, Then check if anything else changes?

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/OssRulesOfPlayScore.java
```diff
@@ -78,7 +80,8 @@
    */
   public static final Set<Feature<Boolean>> RECOMMENDED_TRUE = Collections.unmodifiableSet(
       setOf(HAS_CONTRIBUTING_GUIDELINE, HAS_REQUIRED_TEXT_IN_CONTRIBUTING_GUIDELINE,
-          HAS_CODE_OF_CONDUCT, HAS_REQUIRED_TEXT_IN_CODE_OF_CONDUCT_GUIDELINE));
+          HAS_CODE_OF_CONDUCT, HAS_REQUIRED_TEXT_IN_CODE_OF_CONDUCT_GUIDELINE,
+          RUNS_CODEQL_SCANS,USES_CODEQL_CHECKS));
```
sourabhsparkala, line 84:
```suggestion
          RUNS_CODEQL_SCANS, USES_CODEQL_CHECKS));
```

sourabhsparkala (commented):
Could you also please create an issue and associate this PR to that issue?

src/test/shell/tool/github/OssRulesOfPlayRatingMarkdownFormatter.config.yml
sourabhsparkala: That would not work. Here the rules are being mapped to a specific feature in OssFeatures.java. Please check there for CodeQL, as to what the mapping should be.

docs/getting_oss_rules_of_play_rating.md
ajinkyapatil8190: I tried that locally but nothing changed, 
I got the same score and the same tests were failing on the local system.
The same tests btw are also failing on the upstream project as well.

Sachpat (approved):
Looks good to me. As per my conversation with @ajinkyapatil8190 , the tests that are failing are also failing on the upstream master , so its safe to say they are not caused by the new changes.

Sachpat (approved):
looks good to me, thanks for fixing the tests!

PR #778, opened 2022-02-08
Adding exception for checks for JMC repo

ajinkyapatil8190:
Adding the JMC repository in the exception list for the License and Reuse checks, As it is a fork of openjdk/jmc repository.


src/main/java/com/sap/oss/phosphor/fosstars/data/github/LicenseInfo.java
```diff
@@ -356,7 +356,8 @@ public static void main(String... args) throws Exception {
             + "disallowedLicensePatterns:\n"
             + "  - API\n"
             + "repositoryExceptions:\n"
-            + "  - https://github.com/SAP/SapMachine\n",
+            + "  - https://github.com/SAP/SapMachine\n"
```
Sachpat, line 359: Why don't we maintain this list at one place as some constants or so? So that it can be reused at places like below.  In the future, we just need to modify this list at one place.

PR #822, opened 2022-05-30
Addressing IllegalStateException in OssArtifactSecurityScore

### LINKED ISSUE #820: OssSecurityScore is unknown and OssArtifactSecurityScore is known throws Illegal State Exception
Unavailability of features gives `OssSecurityScore` as `unknown` but when calculating `OssArtifactSecurityScore` an `IllegalStateException` is thrown

```
java.lang.IllegalStateException: It's an unknown value, get() method is not supposed to be called!
	at com.sap.oss.phosphor.fosstars.model.value.ScoreValue.get(ScoreValue.java:144)
	at com.sap.oss.phosphor.fosstars.model.score.oss.OssArtifactSecurityScore.calculate(OssArtifactSecurityScore.java:96)
	at com.sap.oss.phosphor.fosstars.model.score.AbstractScore.calculate(AbstractScore.java:93)
	at com.sap.oss.phosphor.fosstars.model.score.oss.OssArtifactSecurityScoreTest.testCalculateOssSecurityScoreUnknown(OssArtifactSecurityScoreTest.java:66)
```

This exception should not be thrown. The core should rationalize the `OssArtifactSecurityScore` with a `LOW` confidence value.

Things to Do:
- The core should rationalize the `OssArtifactSecurityScore` with a `LOW` confidence value.

sourabhsparkala:
Fixes #820

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/OssArtifactSecurityScore.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/OssArtifactSecurityScore.java
@@ -88,7 +88,9 @@ public class OssArtifactSecurityScore extends AbstractScore {
     }
 
     final double updatedScore;
-    if (artifactVersionSecurityScore.get() >= 9.0) {
+    if (projectSecurityScore.isUnknown()) {
+      updatedScore = artifactVersionSecurityScore.get();
+    } else if (artifactVersionSecurityScore.get() >= 9.0) {
       updatedScore = 0.11 * artifactVersionSecurityScore.get() * projectSecurityScore.get();
     } else {
       updatedScore = 0.1 * (0.9 + artifactVersionSecurityScore.get()) * projectSecurityScore.get();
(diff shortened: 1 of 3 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/OssArtifactSecurityScore.java
```diff
@@ -88,7 +88,9 @@ public ScoreValue calculate(Value<?>... values) {
     }
```
mibo, line 88: I would suggest in case of an unknown score (either artifact or project) to return to combined score as unknown.
Because the `OssArtifactSecurityScore` take the project score and modify it based on the artifact version score.
If the project score is unknown it can not be modified and just returning the artifact version score is not correct (IMHO).

Other opinions?

As sample (which requires changes additional changes in line 91ff):
```suggestion
    } else if (projectSecurityScore.isUnknown()) {
      return scoreValue.withMinConfidence().makeUnknown();
    }
```

mibo (changes requested):

src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/OssArtifactSecurityScore.java
sourabhsparkala: I agree with your argument. The artifact score should be co-dependent with the ProjectScore.

ManjunathMS35 (approved):
LGTM

Issue #817, opened 2022-04-26
Non-Determinism In Unit Test


jchen8460:
There is non-deterministic behavior in the test `com.sap.oss.phosphor.fosstars.data.artifact.VulnerabilitiesFromOwaspDependencyCheckTest.testVulnerabilitiesAvailable`. It comes from assuming a specific order when using the `HashSet` iterator. As stated by the Javadoc, "It makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time." This issue was found by the tool [NonDex](https://github.com/TestingResearchIllinois/NonDex).

I already have a solution for this test. All of the vulnerability id's can be added into a `HashSet`. Then, the specific id can be searched for in the `HashSet`. This ensures that the test will always pass even if the iteration order changes. Please let me know if you want me to open the PR.

sourabhsparkala:
> There is non-deterministic behavior in the test `com.sap.oss.phosphor.fosstars.data.artifact.VulnerabilitiesFromOwaspDependencyCheckTest.testVulnerabilitiesAvailable`. It comes from assuming a specific order when using the `HashSet` iterator. As stated by the Javadoc, "It makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time." This issue was found by the tool [NonDex](https://github.com/TestingResearchIllinois/NonDex).
> 
> I already have a solution for this test. All of the vulnerability id's can be added into a `HashSet`. Then, the specific id can be searched for in the `HashSet`. This ensures that the test will always pass even if the iteration order changes. Please let me know if you want me to open the PR.

@jchen8460 overall it sounds like a good idea. Now that I look at it, its just lazy coding on our end. Please go ahead and create a PR on your free time. Thanks for the contribution.

sourabhsparkala:
This issue is resolved with #823

PR #823, opened 2022-05-31
Address Non-Deterministic Behavior in Unit Test

jchen8460:
**Description**
This PR addresses the issue described here: https://github.com/SAP/fosstars-rating-core/issues/817. 
Essentially, the problem lies in assuming a specific iteration order for a `HashSet` in the test `com.sap.oss.phosphor.fosstars.data.artifact.VulnerabilitiesFromOwaspDependencyCheckTest.testVulnerabilitiesAvailable`. 

**Solution**
We can add all of the vulnerability ids into a `HashSet`. Then, we can just assert that this `HashSet` contains the specific vulnerability id. This way, we make no assumptions about iteration order.

### DIFF
--- a/src/test/java/com/sap/oss/phosphor/fosstars/data/artifact/VulnerabilitiesFromOwaspDependencyCheckTest.java
+++ b/src/test/java/com/sap/oss/phosphor/fosstars/data/artifact/VulnerabilitiesFromOwaspDependencyCheckTest.java
@@ -11,11 +11,12 @@ import com.sap.oss.phosphor.fosstars.data.owasp.model.OwaspDependencyCheckEntry;
 import com.sap.oss.phosphor.fosstars.model.subject.oss.MavenArtifact;
 import com.sap.oss.phosphor.fosstars.model.value.ValueHashSet;
 import com.sap.oss.phosphor.fosstars.model.value.Vulnerabilities;
-import com.sap.oss.phosphor.fosstars.model.value.Vulnerability;
 import com.sap.oss.phosphor.fosstars.util.Json;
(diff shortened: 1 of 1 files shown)


src/test/java/com/sap/oss/phosphor/fosstars/data/artifact/VulnerabilitiesFromOwaspDependencyCheckTest.java
```diff
@@ -49,8 +50,9 @@ public void testVulnerabilitiesAvailable() throws IOException {
     Vulnerabilities vulnerabilities = values.of(VULNERABILITIES_IN_ARTIFACT).get().get();
     assertEquals(3, vulnerabilities.size());
 
-    Vulnerability vulnerability = vulnerabilities.entries().iterator().next();
-    assertEquals("CVE-2018-11307", vulnerability.id());
+    Set<String> vulnerabilityIds = new HashSet<>();
+    vulnerabilities.entries().forEach((vulnerability) -> vulnerabilityIds.add(vulnerability.id()));
+    assertTrue(vulnerabilityIds.contains("CVE-2018-11307"));
```
sourabhsparkala, line 55: I am thinking this can be shortened with help of stream(), as `vulnerabilities` is already a `Set`
```suggestion
    assertTrue(vulnerabilities.entries().stream()
        .anyMatch(vulnerability -> "CVE-2018-11307".equals(vulnerability.id())));
```

sourabhsparkala (changes requested):
@jchen8460 please let me know your thoughts on the review.

PR #837, opened 2022-06-09
ArtifactVersion not set due to incomplete of release list information

### LINKED ISSUE #836: ArtifactVersion not set due to incomplete of release list information
When we want to calculate fosstars-ratings for GAV `com.fasterxml.jackson.core:jackson-databind:2.9.8`
We collect the list of versions for the GA using this URL https://search.maven.org/solrsearch/select?q=g:com.fasterxml.jackson.core+AND+a:jackson-databind&core=gav&wt=json&rows=10000

But this URL doesn't seem to work anymore, that is not all versions are being listed in this URL now. For example version `2.9.8`
Due to this the [block](https://github.com/SAP/fosstars-rating-core/blob/master/src/main/java/com/sap/oss/phosphor/fosstars/data/AbstractReleaseInfoLoader.java#L120-L125) here where checks if the requested version `2.9.8` is present in the list of releases, then set the Artifact version feature to requested version `2.9.8`

But since the list of released versions itself is incomplete, the Artifact Version feature is `unknown`

sourabhsparkala:
Fixes #836


src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
```diff
@@ -35,49 +36,86 @@ public class ReleaseInfoFromMaven extends AbstractReleaseInfoLoader {
...
-      Set<ArtifactVersion> artifactVersions = new HashSet<>();
       for (JsonNode item : docs) {
         if (item.has("v") && item.has("timestamp")) {
           artifactVersions.add(new ArtifactVersion(item.get("v").asText(),
               convertEpochToLocalDate(item.get("timestamp").asLong())));
         }
       }
+    }
+
+    int numFound = json.at("/response/numFound").asInt();
```
Sachpat, line 79: null check on json.at("/response/numFound") would be safer here.

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
```diff
@@ -35,49 +36,86 @@ public class ReleaseInfoFromMaven extends AbstractReleaseInfoLoader {
...
+    if (numFound > artifactVersions.size() && numFound > start) {
```
Sachpat, line 80: logically speaking, shouldn't umFound > artifactVersions.size()  be enough?

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
sourabhsparkala: It would not be necessary, as by default asInt() method will send a `0`

image

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
sourabhsparkala: You are right, but I wanted a fail safe here
For example `numFound > start` will make sure that Pagination does not continue further. Make sure to avoid an infinite recursive call.

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
Sachpat: No you are calling asInt on  json.at("/response/numFound"),  json.at("/response/numFound") can be null

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
Sachpat: Why will the pagination continue if numFound > artifactVersions.size()  ?

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
sourabhsparkala: I have already put in the test case to cover this https://github.com/SAP/fosstars-rating-core/pull/837/files#diff-b347ea38780d8fc3d0ffba6abf5d8f366385933eb84c5b466f84a31c0003fba3R137

Sachpat: No its not about the test case. Will it not throw NPE if json.at("/response/numFound") is null?

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
sourabhsparkala: Yes you are right, without that condition also, all the tests are passing. I am guessing various scenarios are covered.

src/main/java/com/sap/oss/phosphor/fosstars/data/artifact/ReleaseInfoFromMaven.java
sourabhsparkala: It doesn't throw NPE, if `numFound` is `null` or Not found. The test case confirms that.

Sachpat (approved):
Looks good to me.

sourabhsparkala:
@Sachpat thank you for the review.

PR #842, opened 2022-06-30
adding readme.adoc to the readme checks

ajinkyapatil8190:
Fosstars was not checking for readme.adoc file which it should - so adding the code to check for readme.adoc and tests.


sourabhsparkala:
@ajinkyapatil8190 I would suggest you to use a common static method which has the duplicate code and pass the arguments one at a time. Is this is possible. Thanks

src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReadmeInfoTest.java
```diff
@@ -140,6 +140,30 @@ public void testLowercaseReadme() throws IOException {
     assertTrue(checkValue(values, HAS_README, true).get());
   }
 
+  void readmMeAdocTest(String fileName) throws IOException {
```
SebastianWolf-SAP, line 143: That's a generic method, not specific to adoc

src/test/java/com/sap/oss/phosphor/fosstars/data/github/ReadmeInfoTest.java
```diff
@@ -140,6 +140,30 @@ public void testLowercaseReadme() throws IOException {
...
+    when(localRepository.readTextFrom(fileName))
+        .thenReturn(Optional.of(String.join("\n",
+            "This is " + fileName
+        )));
+    ValueSet values = provider.fetchValuesFor(project);
+    assertTrue(checkValue(values, HAS_README, true).get());
+  }
+
+  @Test
+  public void testLowercaseReadmeAdoc() throws IOException {
```
SebastianWolf-SAP, line 162: This tests both scenarios - lowercase and uppercase

SebastianWolf-SAP (commented):
Thanks for the changes and also for including it in tests. I'd say if you include a generic test method, you should name it correctly and also change the other test methods accordingly to use that as well. But in general no blocker for merging.

sourabhsparkala (approved):
👍

PR #850, opened 2022-08-02
Implement new feature Has executable binaries in project

### LINKED ISSUE #733: PoC: Presence of executable binaries
If a binary is present in the GitHub project, there are chances that users would be able to execute it directly. Hence the presence of a binary file would be a security issue.

Parse thorugh the projects to get files of extensions (Ex: `.class` for java. `.pyc` for python), which are executables.

DoD:

- Check if a feature can be extracted.
- A score can be implemented.

sourabhsparkala:
Fixes #733

### DIFF
--- a/src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
+++ b/src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
@@ -1,0 +1,98 @@
+package com.sap.oss.phosphor.fosstars.data.github;
+
+import static com.sap.oss.phosphor.fosstars.model.feature.oss.OssFeatures.HAS_EXECUTABLE_BINARIES;
+
+import com.google.common.base.Predicate;
+import com.sap.oss.phosphor.fosstars.model.Feature;
+import com.sap.oss.phosphor.fosstars.model.Value;
+import com.sap.oss.phosphor.fosstars.model.subject.oss.GitHubProject;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import org.kohsuke.github.GitHub;
+import org.kohsuke.github.GitHubBuilder;
+
+/**
+ * The data provider tries to figure out if an open-source project has executable binaries (for
+ * example, .class, .pyc, .exe)..
+ */
+public class HasExecutableBinaries extends CachedSingleFeatureGitHubDataProvider<Boolean> {
+
+  /**
+   * List of file extensions deemed as executable binaries.
+   */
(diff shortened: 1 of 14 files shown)


src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
```diff
@@ -0,0 +1,98 @@
...
+  @Override
+  protected Value<Boolean> fetchValueFor(GitHubProject project) throws IOException {
+    logger.info("Figuring out if the project has executable binaries...");
+
+    LocalRepository localRepository = loadLocalRepository(project);
+    List<Path> paths = localRepository.files(FILE_EXTENSIONS_PREDICATE);
+    if (paths.size() > 0) {
+      return HAS_EXECUTABLE_BINARIES.value(true);
+    }
+    return HAS_EXECUTABLE_BINARIES.value(false);
```
ManjunathMS35, line 59: Can we simplify this to
    return !HAS_EXECUTABLE_BINARIES.isEmpty();
or return HAS_EXECUTABLE_BINARIES.size() > 0;

src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
```diff
@@ -0,0 +1,98 @@
...
+    return FILE_EXTENSIONS.stream().anyMatch(ext -> path.getFileName().toString().endsWith(ext));
+  }
+
+  /**
+   * This is for testing and demo purposes.
+   *
+   * @param args Command-line options (option 1: API token, option 2: project URL).
+   * @throws Exception If something went wrong.
+   */
+  public static void main(String... args) throws Exception {
```
ManjunathMS35, line 89: This needs to be removed

src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
sourabhsparkala: Yes, we can. But this way the if condition explains better what is happening. I guess it is a personal preference.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/HasExecutableBinaries.java
sourabhsparkala: It can me removed, but it is fine if we leave it here as in the future if someone wants to enhance or provide a fix, they could this config to run the data provider. Atleast that was my intent, we can remove this if you are completely against it.

ManjunathMS35: Does all the data providers have this?

ManjunathMS35 (approved):
LGTM 👍

PR #866, opened 2022-09-01
Static analysis tool providers from Prospector Project 

### LINKED ISSUE #730: PoC: Prospector in Python projects
Prospector is a tool to analyze Python code and output information about errors, potential problems, convention violations and complexity.

- Use of pre-commit config.yml file https://github.com/PyCQA/prospector#pre-commit

DoD:
- Check if the above feature can be implemented as part of rating-core
- Check which score is suitable to use the feature.

sourabhsparkala:
Fixes #730


src/main/java/com/sap/oss/phosphor/fosstars/model/score/oss/MyPyScore.java
```diff
@@ -0,0 +1,88 @@
...
+      return scoreValue.makeUnknown().explain(
+          "The score value is unknown because all required features are unknown.");
+    }
+
+    if (languages.isUnknown()) {
+      return scoreValue.makeNotApplicable().explain(
+          "The score is N/A because the project does not confirm which languages are used.");
+    }
+
+    if (!languages.isUnknown() && !SUPPORTED_LANGUAGES.containsAnyOf(languages.get())) {
```
Sachpat, line 73: !languages.isUnknown()

This is not needed right? Anyway the code will reach here only when !languages.isUnknown() :)

src/main/java/com/sap/oss/phosphor/fosstars/data/AbstractStaticScanToolsDataProvider.java
```diff
@@ -1,37 +1,103 @@
...
+  private static final String GITHUB_INI_CONFIG_EXTENSION = ".ini";
+
+  /**
+   * A Bi-Predicate to evaluate each context of Map type extracted from the GitHub Action jobs.
+   */
+  private static final BiPredicate<Map<?, ?>, Map<String, Predicate<String>>> MATCH_MAP_PREDICATE =
+      (context, matchers) -> {
+        for (Map.Entry<String, Predicate<String>> entry : matchers.entrySet()) {
+          final String key = entry.getKey();
+          if (context.containsKey(entry.getKey())
```
ManjunathMS35, line 74:
```suggestion
          if (context.containsKey(key)
```

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
```diff
@@ -10,31 +10,36 @@
...
+ *  <li>{@link OssFeatures#USES_BANDIT_SCAN_CHECKS}</li>
  * </ul>
  */
 public class BanditDataProvider extends AbstractStaticScanToolsDataProvider {
 
   /**
-   * A step in a GitHub action that triggers analysis with Bandit.
+   * A Predicate to check the any step in a GitHub action that triggers analysis with Bandit.
    */
-  private static final Pattern RUN_STEP_BANDIT_REGEX_PATTERN
```
ManjunathMS35, line 36: Isn't it nice to keep this and use it in line 38 and 40 rather than creating Pattern twice.

src/main/java/com/sap/oss/phosphor/fosstars/data/github/GoSecDataProvider.java
```diff
@@ -33,20 +28,38 @@
...
 public class GoSecDataProvider extends AbstractStaticScanToolsDataProvider {
-
-  /** 
-    * A step in a GitHub action that triggers analysis with GoSec.
-    */
-  private static final Pattern RUN_STEP_GOSEC_REGEX_PATTERN =
-      Pattern.compile("^.*securego/gosec.*$", Pattern.DOTALL);
-
-  /** A step config of a GitHub action that triggers GoSec scans with include rule. */
-  private static final Pattern RUN_STEP_GOSEC_WITH_INCLUDE_REGEX_PATTERN =
```
ManjunathMS35, line 44: Same as above isn't it nice to create these patterns once and use it in the class?

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: There is a slight difference in these regular patterns, the key to search for in the step and then matching the value of the key. But again you are right about the redundancy, i did not find a better way to integrate one predicate and check for everything.

ManjunathMS35 (approved):
Except the minor comments, rest LGTM 👍

src/main/java/com/sap/oss/phosphor/fosstars/data/github/BanditDataProvider.java
sourabhsparkala: @ManjunathMS35 lets fix this here #873 ?

PR #892, opened 2022-10-20
addition of readme file type

ajinkyapatil8190:
As per the issue - https://github.com/sap-tutorials/Tutorials/issues/18895
The repository has the README.MD but fosstars does not understand it, so adding the new file type.


SebastianWolf-SAP (approved):
OK, but maybe we ned to find a better way to cover all cases, this is getting a little bit too much for a list... ;)

PR #959, opened 2023-07-06
Update maven artifacts

ajinkyapatil8190:


sourabhsparkala:
@ajinkyapatil8190 was there a specific reason for this update?

